PR #23, opened 2019-11-05
Bug 1593459 - update to use the community-tc Taskcluster deployment

djmitche:


.taskcluster.yml
```diff
@@ -31,8 +31,8 @@ tasks:
       - taskId: {$eval: as_slugid("lint_test_task")}
         created: {$fromNow: ''}
         deadline: {$fromNow: '1 hour'}
-        provisionerId: aws-provisioner-v1
-        workerType: github-worker
+        provisionerId: proj-relman
```
djmitche: We could change this to `proj-misc` if that project is more appropriate.

marco-c: This project doesn't need secrets yet, but it will (for releases), so proj-relman is better.

PR #91, opened 2020-02-24
Add a way to extract function spaces as json files

Luni-4:
This PR adds a way to extract function spaces as `json` files. It can work both on a single file and on a directory.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -27,6 +27,7 @@ struct Config {
     count_filter: Vec<String>,
     function: bool,
     metrics: bool,
+    output: String,
     line_start: Option<usize>,
     line_end: Option<usize>,
     preproc_lock: Option<Arc<Mutex<PreprocResults>>>,
@@ -72,7 +73,14 @@ fn act_on_file(language: LANG, path: PathBuf, cfg: Config) -> std::io::Result<()
     } else if cfg.metrics {
         let source = read_file_with_eol(&path)?;
         let language = guess_language(&source).unwrap_or(language);
-        let cfg = MetricsCfg { path };
+        let cfg = MetricsCfg {
+            path,
+            output_path: if cfg.output.is_empty() {
+                None
+            } else {
+                Some(PathBuf::from(cfg.output))
+            },
+        };
         action::<Metrics>(&language, source, &cfg.path.clone(), pr, cfg)
     } else if cfg.comments {
         let source = read_file_with_eol(&path)?;
@@ -320,7 +328,7 @@ fn main() {
         )
         .arg(
             Arg::with_name("output")
-                .help("Output file")
+                .help("Output file/directory")
                 .short("o")
                 .long("output")
                 .default_value("")
@@ -439,7 +447,12 @@ fn main() {
         (None, None)
     };
 
-    let output = matches.value_of("output").unwrap();
+    let output = matches.value_of("output").unwrap().to_string();
+    let output_is_dir = PathBuf::from(output.clone()).is_dir();
+    if metrics && !output.is_empty() && !output_is_dir {
+        eprintln!("Error: The output parameter must be a directory");
+        process::exit(1);
+    }
     let language = if preproc_lock.is_some() {
         Some(LANG::Preproc)
     } else if typ.is_empty() {
@@ -472,6 +485,7 @@ fn main() {
         count_filter,
         function,
         metrics,
+        output: output.clone(),
         line_start,
         line_end,
         preproc_lock: preproc_lock.clone(),
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -1,3 +1,4 @@
+use regex::Regex;
 use serde::ser::{SerializeStruct, Serializer};
 use serde::Serialize;
 use std::fmt;
@@ -14,6 +15,7 @@ use crate::fn_args::{self, NArgs};
 use crate::getter::Getter;
 use crate::halstead::{self, Halstead};
 use crate::sloc::{self, SourceLoc};
+use crate::tools::write_file;
 use crate::traits::*;
 
 #[derive(Debug)]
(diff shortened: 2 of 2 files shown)


calixteman:
This patch looks pretty good :) but there is already an option "-o" to set the output so I'd say that "--metrics -o foo" should dump the json in foo.
We should probably dump on stdout or in a file according to "-o" presence:
https://github.com/mozilla/rust-code-analysis/blob/master/src/metrics.rs#L396

Luni-4:
> This patch looks pretty good :) but there is already an option "-o" to set the output so I'd say that "--metrics -o foo" should dump the json in foo.

I tried to run the following command: `./rust-code-analysis --metrics -o foo -p file.rs`

But it doesn't create the `foo` file since the file creation is related to the preproc part as you can see from [here](https://github.com/mozilla/rust-code-analysis/blob/master/src/main.rs#L531)

> We should probably dump on stdout or in a file according to "-o" presence:
> https://github.com/mozilla/rust-code-analysis/blob/master/src/metrics.rs#L396

Good idea! We could add a `json_bool` field to the `MetricsCfg` struct and move `dump_json` into `FuncSpace`. But what about the `-o` option? It works differently as it is now.

calixteman:
"-o" isn't especially related to "-preproc".
What I proposed to you is just to modify your patch to get the "-o" value, pass it to Config and when used with metrics=true, then pass it into MetricsCfg... and then if we've an output path in MetricsCfg then dump the json in it else dump on stdout.

Luni-4:
> "-o" isn't especially related to "-preproc".

I see, thanks for your explanation!

> What I proposed to you is just to modify your patch to get the "-o" value, pass it to Config and when used with metrics=true, then pass it into MetricsCfg... and then if we've an output path in MetricsCfg then dump the json in it else dump on stdout.

Yep, I'm going to do it, thanks!

src/main.rs
```diff
@@ -72,8 +73,20 @@ fn act_on_file(language: LANG, path: PathBuf, cfg: Config) -> std::io::Result<()
     } else if cfg.metrics {
         let source = read_file_with_eol(&path)?;
         let language = guess_language(&source).unwrap_or(language);
-        let cfg = MetricsCfg { path };
-        action::<Metrics>(&language, source, &cfg.path.clone(), pr, cfg)
+        if cfg.output.is_empty() {
```
calixteman: nit: there is some dup code in then & else clauses (finally the only difference is almost output_path definition in the struct)

src/main.rs
```diff
@@ -320,7 +333,7 @@ fn main() {
         )
         .arg(
             Arg::with_name("output")
-                .help("Output file")
+                .help("Output directory")
```
calixteman: I'd say `file/directory`.
`file` was too restrictive but `directory` is too

src/main.rs
```diff
@@ -439,7 +452,12 @@ fn main() {
         (None, None)
     };
 
-    let output = matches.value_of("output").unwrap();
+    let output = matches.value_of("output").unwrap().to_string();
+    let output_is_dir = PathBuf::from(output.clone()).is_dir();
+    if !output.is_empty() && !output_is_dir {
```
calixteman: We can adapt according to the kind of path we've and the tool we're using: I mean `--metrics -o not_dir` is forbidden but can be allowed with `--foo -o not_dir`.
So just disallow file type when we're running metrics.

src/metrics.rs
```diff
@@ -306,6 +307,17 @@ impl<'a> FuncSpace<'a> {
         color!(stdout, White);
         writeln!(stdout, "{}", val)
     }
+
+    fn dump_json(&self, path: &PathBuf, output_path: &PathBuf) -> std::io::Result<()> {
+        let json_data = serde_json::to_string_pretty(&self).unwrap();
```
calixteman: Could you file a bug to add an option `--pretty` to choose if the dump should be pretty or not ?

src/main.rs
Luni-4: You're right, I will change that, thanks!

src/metrics.rs
Luni-4: Sure! :)

Luni-4: Done

src/main.rs
Luni-4: Fixed

src/metrics.rs
```diff
@@ -306,6 +307,17 @@ impl<'a> FuncSpace<'a> {
...
+
+        let mut file = path.as_path().file_name().unwrap().to_os_string();
+        file.push(".json");
+
+        let mut json_path = output_path.clone();
+        json_path.push(file);
+        write_file(&json_path, json_data.to_string().as_bytes())
```
calixteman: nit: the `to_string` is probably useless

calixteman: how to deal with an existing file ?
For example, you have `./foo/a.py` and `./bar/a.py`, the two files `a.py.json` will be the same.

Luni-4: We could look if there is already an `a.py.json` file in the `output_path` directory and replace filenames with paths (e.g `./foo/a.py.json` and ./bar/a.py.json`) 

What do you think of this approach?

calixteman: It's ok for me, let's do that and if someone complains at some point we'll add an option or whatever...

Luni-4: > It's ok for me, let's do that and if someone complains at some point we'll add an option or whatever...

I agree

> FYI, @mozilla we use the server part for now.

I see, that's why is full of tests, thanks for the info! :)

Luni-4: @calixteman 
Done. I've replaced slashes, backslashes and colons with `_`.

Luni-4:
@calixteman 
Is still there something I need to do such that this PR could be landed?

calixteman:
Grazie mille

Luni-4:
Grazie a te :)

PR #102, opened 2020-03-05
Unbreak macOS

lu-zero:
macOS ar expects to build non-empty archives.


sylvestre:
thanks for your contribution!

PR #105, opened 2020-03-05
Add an example

sylvestre:


sylvestre:
Calixte told me that he prefers to highlight CLI usage

PR #110, opened 2020-03-12
Implement cloc metric

Luni-4:
- Implemented `Cloc` metric
- Updated tests

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -102,13 +102,23 @@ impl<'a> Serialize for FuncSpace<'a> {
 
 impl<'a> FuncSpace<'a> {
     fn new<T: Getter>(node: &Node<'a>, code: &'a [u8], kind: NodeKind) -> Self {
+        let (start_position, end_position) = match kind {
+            NodeKind::Unit => {
+                if node.child_count() == 0 {
+                    (0, 0)
+                } else {
+                    (node.start_position().row + 1, node.end_position().row)
+                }
+            }
+            _ => (node.start_position().row + 1, node.end_position().row + 1),
+        };
         Self {
             name: T::get_func_space_name(&node, code),
             spaces: Vec::new(),
             metrics: CodeMetrics::default(),
             kind,
-            start_line: node.start_position().row + 1,
-            end_line: node.end_position().row + 1,
+            start_line: start_position,
+            end_line: end_position,
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/metrics.rs
```diff
@@ -102,13 +102,24 @@ impl<'a> Serialize for FuncSpace<'a> {
 
 impl<'a> FuncSpace<'a> {
     fn new<T: Getter>(node: &Node<'a>, code: &'a [u8], kind: NodeKind) -> Self {
+        let mut start_position = node.start_position().row;
```
calixteman: I'm not happy this change because start_line & end_line have not their expected values or maybe I missed something.
Suppose you have a function like `fn foo() -> u32 { 1 }`, with your change it's location (i.e. start/end_line) will be 0, 0.

src/sloc.rs
```diff
@@ -146,14 +172,14 @@ impl SourceLoc for TsxCode {
...
+    fn compute(node: &Node, _code: &[u8], stats: &mut Stats, is_func_space: bool, is_unit: bool) {
         use Rust::*;
 
-        let start = init(node, stats, is_func_space);
+        let start = init(node, stats, is_func_space, is_unit);
 
         match node.kind_id().into() {
             LineComment | BlockComment | StringLiteral | RawStringLiteral | ExpressionStatement
-            | Block => {}
+            | Block | SourceFile => {}
```
calixteman: good catch

src/metrics.rs
Luni-4: Oops, you're completely right, it should be fixed now, thanks a lot!

src/sloc.rs
Luni-4: Thank you! :)

src/metrics.rs
calixteman: You could probably just check if the unit is empty (i.e. no children) and then set position to 0 in this only case, I suppose in this case it's acceptable.

Luni-4: > You could probably just check if the unit is empty (i.e. no children) and then set position to 0 in this only case, I suppose in this case it's acceptable.

Okay, I'm going to change my implementation, thanks!

> FYI, I always inject an EOL as last char: https://github.com/mozilla/rust-code-analysis/blob/master/src/tools.rs#L39 because some parsers expect to have one.

Now I know because I got (1,1) when I tried with an empty file. But is it always right to add an `EOL` at the end of a file? Because a user, not knowing of this feature, could be confused imho.

calixteman: The EOL injection is just a workaround (only when no last eol).
In all firefox source code, I found only one cpp file with this final eol and in this case the parsing was kinda broken.

Luni-4: I see, thanks, so I will implement what you said above without changing anything else

Luni-4: Done

src/sloc.rs
```diff
@@ -41,40 +49,58 @@ impl Stats {
 
     #[inline(always)]
     pub fn sloc(&self) -> f64 {
-        (self.end - self.start) as f64 + 1.
+        let sloc = if self.unit {
+            self.end - self.start
```
calixteman: ncould you add a comment to explain why there is a difference here in the way to compute sloc ?

Luni-4: Ok, I've added a comment.

PR #112, opened 2020-03-14
Add an additional CI system

Luni-4:
In order to test `rust-code-analysis` on different operating systems, I've added `GitHub Actions`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -105,7 +105,7 @@ fn build_c(files: Vec<String>, language: &str) {
             .opt_level(get_opt_level())
             .debug(get_debug())
             .warnings(false)
-            .flag("-std=c99");
+            .flag_if_supported("-std=c99");
     }
     build.compile(&format!("tree-sitter-{}-c", language));
 }
--- a/tree-sitter-mozcpp/src/scanner.cc
+++ b/tree-sitter-mozcpp/src/scanner.cc
@@ -1,1 +1,1 @@
-../../tree-sitter-cpp/src/scanner.cc
+#include "../../tree-sitter-cpp/src/scanner.cc"
--- a/.github/workflows/rust-code-analysis.yml
+++ b/.github/workflows/rust-code-analysis.yml
@@ -1,0 +1,40 @@
+name: Run tests
+
+on: [push, pull_request]
+
+jobs:
+  test-code:
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Checkout submodules
+      shell: bash
+      run: |
+        GIT_PATH=https://github.com/.extraheader
+        AUTH_HEADER="$(git config --local --get http.$GIT_PATH)"
+        git submodule sync --recursive
+        git -c "http.extraheader=$AUTH_HEADER" \
+            -c protocol.version=2 \
+            submodule update --init --force --recursive --depth=1
+
+    - name: Install Rust stable
+      uses: actions-rs/toolchain@v1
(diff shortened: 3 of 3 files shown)


marco-c:
Unfortunately I think we can't use GitHub Actions in the mozilla GitHub org for the time being (https://bugzilla.mozilla.org/show_bug.cgi?id=1594423).
We can probably support testing on other operating systems via Taskcluster, I will figure out how.

Luni-4:
Oh, I see. If we can use `Taskcluster` to test other operating systems, it would be fine for me.

marco-c:
It should be possible to use Windows after https://github.com/mozilla/community-tc-config/pull/238, here's an example: https://github.com/taskcluster/taskcluster-worker-runner/blob/2ebe797d89f3dd3c814ea4f881a2a5c6306653db/.taskcluster.yml#L76-L113.

Luni-4:
> It should be possible to use Windows after [mozilla/community-tc-config#238](https://github.com/mozilla/community-tc-config/pull/238), here's an example: https://github.com/taskcluster/taskcluster-worker-runner/blob/2ebe797d89f3dd3c814ea4f881a2a5c6306653db/.taskcluster.yml#L76-L113.

Good, thanks! Is Visual Studio already installed? Because if it's not, it will be difficult to install that every time

marco-c:
I think we have all we need, this is the setup script: https://raw.githubusercontent.com/taskcluster/generic-worker/2509a1f8589b0b536186f5fe8f043385bf67197f/worker_types/win2012r2/bootstrap.ps1. It is also used to create the image needed to build Firefox, so a lot of things are already available.

Luni-4:
Perfect Let me know when the `TaskCluster` solution will be ready :)

marco-c:
https://github.com/mozilla/community-tc-config/pull/238 has been merged, we should now be able to use it.

Luni-4:
> [mozilla/community-tc-config#238](https://github.com/mozilla/community-tc-config/pull/238) has been merged, we should now be able to use it.

Great! So are you working on that @marco-c ?

build.rs
```diff
@@ -104,8 +105,11 @@ fn build_c(files: Vec<String>, language: &str) {
             .pic(true)
             .opt_level(get_opt_level())
             .debug(get_debug())
-            .warnings(false)
-            .flag("-std=c99");
+            .warnings(false);
+
+        if triple != "x86_64-pc-windows-msvc" {
```
calixteman: nit: you should use `flag_if_supported`

tree-sitter-mozcpp/src/scanner.cc
```diff
@@ -1 +0,0 @@
-../../tree-sitter-cpp/src/scanner.cc
```
calixteman: Why don't you keep a symbolic link here ?
Removing it means that we need to manually sync this file coming from tree-sitter-cpp.

Luni-4: It didn't work on `Windows`, perhaps I could fix that using the `bash` shell instead of `Powershell`.

calixteman: good idea :) 
simple and easy (I was thinking about a more complex solution)

Luni-4: Thanks to you for finding my initial horrible implementation! :)

marco-c:
> Great! So are you working on that @marco-c ?

I started a WIP PR just to show the skeleton, not sure when I'll be able to continue it though.
Feel free to take it over if you want.

I'm closing this since we can't use GitHub Actions, feel free to open a new one if you want to take #122 over.

Luni-4:
Perfect! Thanks a lot, @marco-c, I will try to update your PR then.

PR #114, opened 2020-03-17
Rename sloc metric suite to loc

Luni-4:
To be more consistent with the scientific literature, the entire set of metrics has been renamed from `sloc` to `loc`.


Luni-4:
@calixteman 
Now all tests pass :)

PR #115, opened 2020-03-17
Update Halstead

Luni-4:
- Add two more metrics to the `Halstead `suite
- Use notations present in the scientific literature
- Update `Halstead` dump
- Update tests


src/halstead.rs
```diff
@@ -287,16 +330,16 @@ impl Halstead for CppCode {
         let id = node.kind_id();
 
         match id.into() {
-            DOT | LPAREN | COMMA | STAR | GTGT | COLON | Return | Break | Continue | If | Else
-            | Switch | Case | Default | For | While | Goto | Do | Delete | New | Try | Catch
-            | Throw | EQ | AMPAMP | PIPEPIPE | PLUS | PLUSPLUS | SLASH | PERCENT | PIPE | AMP
-            | LTLT | TILDE | LT | LTEQ | EQEQ | BANGEQ | GTEQ | GT | PLUSEQ | BANG | STAREQ
-            | SLASHEQ | PERCENTEQ | GTGTEQ | LTLTEQ | AMPEQ | CARET | CARETEQ | PIPEEQ | LBRACK
-            | LBRACE | QMARK | COLONCOLON | TypeSpecifier | Sizeof => {
+            DOT | LPAREN2 | COMMA | STAR | GTGT | COLON | SEMI | Return | Break | Continue | If
```
calixteman: Why do you remove LPAREN ?
Anyway I just checked the diff between LPAREN and LPAREN2, LPAREN is the parenthesis used in macro definitions (e.g. `#define A(...)` and LPAREN2 in function definition or function call.

Luni-4: Oh, my bad. I will re-add that immediately

src/halstead.rs
```diff
@@ -19,13 +19,15 @@ impl Serialize for Stats<'_> {
     where
         S: Serializer,
     {
-        let mut st = serializer.serialize_struct("halstead", 12)?;
-        st.serialize_field("unique_operands", &self.u_operands())?;
-        st.serialize_field("operands", &self.operands())?;
-        st.serialize_field("unique_operators", &self.u_operators())?;
-        st.serialize_field("operators", &self.operators())?;
+        let mut st = serializer.serialize_struct("halstead", 14)?;
+        st.serialize_field("n1", &self.u_operators())?;
```
calixteman: I don't care so much about the names... so why not.
We use them here: https://github.com/mozilla/bugbug/blob/master/bugbug/repository.py#L143 (and not directly here https://github.com/mozilla/bugbug/blob/master/bugbug/commit_features.py#L134)
So if @marco is ok with this change, you should make a PR on bugbug to update the names too.

marco-c: It works for me!

Luni-4: Perfect, I will update the names on those repositories too.

marco-c: NOTE: we'll be able to update them after releasing a new version of rust-code-analysis (which we can do right after merging this PR). In bugbug we'll have to update the names in repository.py like @calixteman mentioned, and update the rust-code-analysis version used as done in https://github.com/mozilla/bugbug/commit/3bb5ab4ca4e76d7f0785821ce0cf711722ee1c40.

Luni-4: I see, so let's merge this PR and eventually update those repositories. I agree with you.

PR #121, opened 2020-03-18
Improve language detection and return used language in the metrics

calixteman:


src/languages.rs
```diff
@@ -48,33 +53,61 @@ mk_langs!(
...
+    ),
     (
         Cpp,
+        "c/c++",
         CppCode,
         CppParser,
         tree_sitter_cpp,
-        [cpp, cxx, cc, hxx, hpp, c, h, hh, inc],
-        ["c++", "c"]
+        [cpp, cxx, cc, hxx, hpp, c, h, hh, inc, mm],
```
marco-c: `m` too

PR #124, opened 2020-03-20
Add Windows CI

Luni-4:
This PR adds a CI for `Windows`

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -58,11 +58,40 @@ tasks:
           owner: cdenizet@mozilla.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+      - taskId: {$eval: as_slugid("windows_test_task")}
+        created: {$fromNow: ''}
+        deadline: {$fromNow: '1 hour'}
+        provisionerId: proj-relman
+        workerType: win2012r2
+        payload:
+          maxRunTime: 3600
+          command:
+            - set VS_PATH=%programfiles(x86)%\Microsoft Visual Studio 14.0\VC
+            - set RUSTUP_INIT_PATH=%CD%\rustup-init
+            - set PATH=%RUSTUP_INIT_PATH%;%USERPROFILE%\.cargo\bin;%PATH%
+            - call "%VS_PATH%\vcvarsall.bat" x86_amd64
+            - rustup-init -yv --default-toolchain stable ^
+                              --default-host x86_64-pc-windows-msvc
+            - git clone --recursive --quiet ${repository}
+            - cd rust-code-analysis
+            - git -c advice.detachedHead=false checkout ${head_rev}
+            - cargo test --verbose --all-features
+          mounts:
+            - content:
+                url: https://win.rustup.rs/
+              file: rustup-init\rustup-init.exe
+        metadata:
+          name: rust-code-analysis windows test
+          description: rust-code-analysis windows test
+          owner: cdenizet@mozilla.com
+          source: ${repository}/raw/${head_rev}/.taskcluster.yml
+
       - $if: 'tasks_for == "github-push" && head_branch[:10] == "refs/tags/"'
         then:
           taskId: {$eval: as_slugid("build_release")}
           dependencies:
             - {$eval: as_slugid("lint_test_task")}
+            - {$eval: as_slugid("windows_test_task")}
           created: {$fromNow: ''}
           deadline: {$fromNow: '2 hour'}
           provisionerId: proj-relman


calixteman:
According to the log I suppose we're using powershell so maybe you could try:
https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-7

Luni-4:
> According to the log I suppose we're using powershell so maybe you could try:
> https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-7

Thanks a lot! :)
I followed your advice, let's see if it works

Luni-4:
@calixteman @marco-c 

'Invoke-WebRequest' is not recognized as an internal or external command, operable program or batch file.

Am I doing something wrong?

marco-c:
You could try using `mounts` to install Rust (example usage of mounts in https://github.com/taskcluster/taskcluster-worker-runner/blob/2ebe797d89f3dd3c814ea4f881a2a5c6306653db/.taskcluster.yml#L98 or https://github.com/mozilla/DeepSpeech/blob/5e46d702afd0729783bd4be1c78116f8f9864bc9/taskcluster/win-opt-base.tyml#L72).

Another option could be to use `choco` but installing the package in a user directory where you have write permissions.

Yet another option could be to use pacman from msys (installing msys like https://github.com/mozilla/DeepSpeech/blob/5e46d702afd0729783bd4be1c78116f8f9864bc9/taskcluster/win-opt-base.tyml#L73, then using a command such as https://github.com/mozilla/DeepSpeech/blob/5e46d702afd0729783bd4be1c78116f8f9864bc9/taskcluster/win-opt-base.tyml#L60).

.taskcluster.yml
```diff
@@ -58,11 +58,38 @@ tasks:
...
+            - rustup-init -yv --default-toolchain stable --default-host x86_64-pc-windows-msvc
+            - git clone --recursive --quiet ${repository}
+            - cd rust-code-analysis
+            - git -c advice.detachedHead=false checkout ${head_rev}
+            - cargo test --verbose --all-features
+          mounts:
+            - content:
+                url: https://win.rustup.rs/
+              directory: rust-init
+              format: exe
```
marco-c: I think in this case you'd need a `file` mount (see docs at https://docs.taskcluster.net/docs/reference/workers/generic-worker/multiuser-windows-payload).
Something like:
```
- content:
    url: https://win.rustup.rs/
  file: rust-init.exe
```

Luni-4: Oh, thank you! :)

If I want to put the exe inside a directory, do I need to create it or I can just set this one below?

```
- content:
    url: https://win.rustup.rs/
  file: rust-init\rust-init.exe
```

calixteman:
Ahhhh!! patches for CI...

Luni-4:
@calixteman 
Ahhh, you're right, so frustrating...

Practically I can't find the `link.exe` from `Visual Studio`, do you know how to retrieve that @marco-c?

calixteman:
Maybe in runing `vcvarsall.bat`: https://github.com/rust-lang/rust/issues/38584#issuecomment-269215096

Luni-4:
Yep, I've tried to run that, but the path is not correct. Which `Visual Studio` version do we have on `TaskCluster`?

```
set VS_PATH="%ProgramFiles(x86)%\Microsoft Visual Studio\2019"
call "%VS_PATH%\Community\VC\Auxiliary\Build\vcvars64.bat"
```

calixteman:
Maybe 14: https://github.com/mozilla/DeepSpeech/blob/5e46d702afd0729783bd4be1c78116f8f9864bc9/taskcluster/win-opt-base.tyml#L67

Luni-4:
Nope, the path doesn't exist

Luni-4:
It seems there isn't `Visual Studio` installed

marco-c:
This is all we have in the image: https://raw.githubusercontent.com/taskcluster/generic-worker/2509a1f8589b0b536186f5fe8f043385bf67197f/worker_types/win2012r2/bootstrap.ps1.

The Windows SDK 8.1 is installed:
> choco install -y windows-sdk-8.1

It should probably have the linker, we "just" need to figure out how to use it :P

calixteman:
Don't we have a windows image with VS inside ?
I'd say that linker should come with VS (probably 2012 with this image)

marco-c:
This image is used to build Firefox, so the linker should be available.

marco-c:
It could be in some directory under C:\Program Files\Microsoft SDKs or C:\Program Files\Windows Kits

Luni-4:
Would it be possible to install `Visual Studio 2019 `on the image?  Because the main problem is the `tree-sitter` compilation. It will be much simpler to use that imho. 

Anyway, the linker should be installed by this command:

`Start-Process "C:\vcredist_x86-vs2015.exe" -ArgumentList "/install /passive /norestart /log C:\vcredist_x86-vs2015-install.log" -Wait -PassThru`

Now I just need to know where :P

marco-c:
(This is the setup script for the image: https://raw.githubusercontent.com/taskcluster/generic-worker/2509a1f8589b0b536186f5fe8f043385bf67197f/worker_types/win2012r2/bootstrap.ps1).

marco-c:
> Would it be possible to install Visual Studio 2019 on the image?

It'd probably increase the size of the image too much, they just install the SDK and other dependencies because not all features of Visual Studio (especially the UI) are needed.

calixteman:
We need to have a C/C++ compiler for tree-sitter stuff + linker of course.
So I'd say that we need to get VS or maybe we could use a clang artifact.

marco-c:
> We need to have a C/C++ compiler for tree-sitter stuff + linker of course.
So I'd say that we need to get VS or maybe we could use a clang artifact.

I don't think Visual Studio is necessary, the Windows SDK (which we have in the image) should be enough.

calixteman:
> > We need to have a C/C++ compiler for tree-sitter stuff + linker of course.
> > So I'd say that we need to get VS or maybe we could use a clang artifact.
> 
> I don't think Visual Studio is necessary, the Windows SDK (which we have in the image) should be enough.

You're right: https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions/Windows_SDK_versions

marco-c:
I'm trying to get access to a machine with that imageset so I can check where the linker is.

marco-c:
link exe

marco-c:
So it seems the Microsoft linker is not installed with the SDK with Choco :(

We could try with the linker from C:\go\pkg\tool\windows_amd64\, maybe we are lucky.
Otherwise, we'd probably need this package: https://chocolatey.org/packages/microsoft-visual-cpp-build-tools.

Luni-4:
I found a linker in this folder `tmp\usr\bin` with my previous commits, I will try to use that first, then the `Go` one.

> Otherwise, we'd probably need this package: https://chocolatey.org/packages/microsoft-visual-cpp-build-tools.

I've already used that package, it takes quite a lot to be installed, so we should add them to the image and recreate it.

Luni-4:
It doesn't work the `go` path either. 
Could you install the `build-tools` on the image, please, @marco-c? Thanks a lot!

marco-c:
We should start with installing it here, and then we can figure out if we can add it to the image (we might have to define a separate one, I need to talk with who manages that image).
Luckily on a cloud machine the command doesn't take so much: just 1 minute.

calixteman:
Or maybe we could use the deepspeech image which should contain what we need.

marco-c:
Yeah, we could do that and hope they keep what we need. If they don't, we'll have to define our own.
This is the setup script they are using: https://github.com/mozilla/community-tc-config/blob/9e3918d97aa629a300c403a84c20da49a3c7b1d4/imagesets/deepspeech-win2012r2-gpu/bootstrap.ps1.

Luni-4:
Good idea @calixteman! :)

It has what we need [https://github.com/mozilla/community-tc-config/blob/9e3918d97aa629a300c403a84c20da49a3c7b1d4/imagesets/deepspeech-win2012r2-gpu/bootstrap.ps1#L53](https://github.com/mozilla/community-tc-config/blob/9e3918d97aa629a300c403a84c20da49a3c7b1d4/imagesets/deepspeech-win2012r2-gpu/bootstrap.ps1#L53)

calixteman:
There is a lot of useless stuff for us in this image and we should probably create one for "standard" projects so we should create new one.
But at least we can try with the deepspeech one and we'll switch to the future one.

Luni-4:
It seems I can't change the image

marco-c:
Not yet, the right scopes need to be added. I'll do it on Monday.

lu-zero:
> These problems will be solved as soon as the `tree-sitter-core` will be [oxidized](https://github.com/lu-zero/tree-sitter-core)

I'm not sure how upstream would be amenable to this, we could complete a first pass by oxidizing the rust and the cpp parsers and see how they work with r-c-a first.

marco-c:
Let's start by trying to install the package here, once we get it working we can decide about what to do with the image. Updating the image is a long process, so the trial and error to update the image first and then this PR, then the image again, etc., etc., would be super painful.

You could try to install https://chocolatey.org/packages/visualstudio2019buildtools, which can also be installed without admin rights if you pass the `--installPath` parameter (it should be something like `choco install -y visualstudio2019buildtools --params "--installPath C:\SOME_DIR`).

Luni-4:
@marco-c 

I tried your method but I get the following error `Access to the path 'C:\ProgramData\chocolatey\lib-bad' is denied.` since `chocolatey` is not authorized to access to that directory. 

Furthermore, that directory can't be changed with the `--install-dir `argument, because we don't have the licensed version of `chocolatey`. 

Other than that, if we succeed, how can we reboot the machine? From the `visualstudio2019buildtools` package: 

```
A reboot may be required after (or even before) installing/uninstalling this package.
If control over reboots is required, it is advisable to install the dependencies (esp. dotnetfx 4.7.2 or later) first, perform a reboot if necessary, and then install this package.
```

marco-c:
> Furthermore, that directory can't be changed with the --install-dir argument, because we don't have the licensed version of chocolatey.

That's the chocolatey argument, but we can instead pass the installation directory as a package parameter (so it's passed to the VS installer) like I mentioned above:
> it should be something like `choco install -y visualstudio2019buildtools --params "--installPath C:\SOME_DIR"`

> Other than that, if we succeed, how can we reboot the machine? From the visualstudio2019buildtools package:

Let's hope it's not required for us, we'll see.

Luni-4:
> That's the chocolatey argument, but we can instead pass the installation directory as a package parameter (so it's passed to the VS installer) like I mentioned above:

> > it should be something like `choco install -y visualstudio2019buildtools --params "--installPath C:\SOME_DIR"`
 
I tried but it doesn't seem to work because of the error mentioned above: [https://github.com/mozilla/rust-code-analysis/pull/124/commits/c0d78bff4b335aed8280ef5c4e356b2a5359a7b6](https://github.com/mozilla/rust-code-analysis/pull/124/commits/c0d78bff4b335aed8280ef5c4e356b2a5359a7b6)

marco-c:
> I tried but it doesn't seem to work because of the error mentioned above: c0d78bf

OMG, this is super painful. I guess we'll have to do trial and error with the image. I'll comment here when it's ready.

Luni-4:
> > I tried but it doesn't seem to work because of the error mentioned above: [c0d78bf](https://github.com/mozilla/rust-code-analysis/commit/c0d78bff4b335aed8280ef5c4e356b2a5359a7b6)
> 
> OMG, this is super painful. I guess we'll have to do trial and error with the image. I'll comment here when it's ready.

Thank you! :)
If you need some help, ping me.

Perhaps the simplest thing would be to install a licensed version of chocolatey on the image, otherwise the build tools directly.

marco-c:
Image updated, you can use `C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat`.

Luni-4:
> Image updated, you can use `C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat`.

Thanks a lot @marco-c! :)

I will work on that on Monday!

marco-c:
> Thanks a lot @marco-c! :)

No, thank you ;)

marco-c:
\o/

Luni-4:
Yeah!! :D

PR #131, opened 2020-03-26
Implement LLOC metric

Luni-4:
This PR implements the `LLOC` metric, so `rust-code-analysis` can now count logical lines. The current `lloc` is renamed as `ploc` since it counts physical lines.


src/loc.rs
```diff
@@ -220,11 +290,18 @@ impl Loc for CppCode {
...
+            RawStringLiteral | StringLiteral | DeclarationList | FieldDeclarationList
+            | TranslationUnit => {}
             Comment => {
                 stats.comment_lines.insert(start);
             }
+            WhileStatement | SwitchStatement | CaseStatement | IfStatement | ForStatement
+            | ReturnStatement | BreakStatement | ContinueStatement | GotoStatement
+            | ThrowStatement | TryStatement | ExpressionStatement | LabeledStatement
+            | StatementIdentifier => {
+                *stats.logical_lines.entry(start).or_insert(0) += 1;
```
calixteman, line 302: So for example with `for (;;) break;` the lloc will be 2.
Is it what you're expecting ?
If yes please add documentation and a test would be nice too.

Luni-4: Can we consider `;;` as statements? If yes, lloc should be 2.

Anyway, this PR is not ready yet. I need to work on that more (unit tests + other checks)

calixteman: It's already 2 because we've a ForStmt and a BreakStmt

Luni-4: > It's already 2 because we've a ForStmt and a BreakStmt.

True, but can we consider this output correct? 

> So finally this metric is number of statements, correct ?

Exactly

> If it's the case the lloc name seems to be inappropriate.

lloc means `Logical Lines of Code` and according to Wikipedia and other papers, logical lines are statements. That's why I've maintained that name.

Luni-4:
- Change metric name
- Add more comments
- Add unit tests

src/loc.rs
```diff
@@ -65,9 +76,16 @@ impl Stats {
...
     #[inline(always)]
     pub fn lloc(&self) -> f64 {
+        // This metric counts the instruction lines present in the code
         self.lines.len() as f64
     }
 
+    #[inline(always)]
+    pub fn stat(&self) -> f64 {
+        // This metric counts all the statements present in the source code
+        self.logical_lines.values().sum::<u64>() as f64
```
calixteman, line 86: Names are confusing here... maybe something like statements_by_line (`stat` itself looks like stats for statistics)

Luni-4: I'd rather to maintain a known acronym like the other `loc` metrics. What do you think of this mapping?

lloc -> ploc (Physical lines of code)
stat -> lloc (Logical Lines of Code)
 
With some comments to explain the acronyms addressed by this metric.

PR #134, opened 2020-04-02
Implement NOM metric

Luni-4:
`NOM` metric counts the number of functions and closures contained in a file/trait/class


src/nom.rs
```diff
@@ -0,0 +1,201 @@
...
+    fn compute(_node: &Node, _code: &[u8], _stats: &mut Stats) {}
+}
+
+impl Nom for PythonCode {
+    fn compute(node: &Node, _code: &[u8], stats: &mut Stats) {
+        use Python::*;
+
+        match node.kind_id().into() {
+            FunctionDefinition => {
+                stats.functions.insert(std::string::String::new());
```
calixteman, line 81: Why do you insert an empty string here ?
The count wil be 0 or 1.

src/nom.rs
```diff
@@ -0,0 +1,201 @@
...
+        }
+    }
+}
+
+impl Nom for RustCode {
+    fn compute(node: &Node, code: &[u8], stats: &mut Stats) {
+        use Rust::*;
+
+        match node.kind_id().into() {
+            FunctionSignatureItem | FunctionItem => {
```
calixteman, line 160: If we have something like:
```rust
mod A { fn foo() {}}
mod B { fn foo() {}}
```
then when merging mod A (as a func space) and mod B in the unit then you'll have only one `foo` which is wrong

src/nom.rs
```diff
@@ -0,0 +1,201 @@
...
+        }
+    }
+}
+
+impl Nom for CppCode {
+    fn compute(node: &Node, code: &[u8], stats: &mut Stats) {
+        use Cpp::*;
+
+        match node.kind_id().into() {
+            FunctionDeclarator | FunctionDeclarator2 | FunctionDeclarator3 => {
```
calixteman, line 180: Because of polymorphism it's probably wrong here:
```cpp
struct A {
  void foo(int) {}
  void foo(double) {}
};
```
The 2 functions have the same name but are different

src/nom.rs
Luni-4: I've used `String` in the `FxHashSet` instead of numbers because I wanted to avoid this case:

```
trait Foo {
 fn foo();
}

impl Foo for Test {
  fn foo() {
    let a = 0;
  }
}
```

In this case there would be 2 `foo` identifier for the unit space, and that would be wrong because the two `foo` are the same. One is the declaration and one its implementation. 

Since we have this case for `Rust` and not for other languages, I inserted an empty string when a method is encountered.

Luni-4:
How can we consider definition and implementation? Should I count one if I declare and implement the same `foo` method in the same file or two?

src/nom.rs
calixteman: If you run rca -m on:
```python
def a():
    pass

def b():
    pass

def c():
    pass
```
you'll get nom::functions=1 which is wrong, correct ?

calixteman: And if I dump the ast for your rust example (`rca -d`), we can see that the parent node for `fn foo();` is a `function_signature_item` and for the other `foo` it's a `function_item`. So you should be able to discriminate definition vs declaration using the context.

calixteman:
> How can we consider definition and implementation? Should I count one if I declare and implement the same `foo` method in the same file or two?

I think we should only consider raw count of definitions.

src/nom.rs
Luni-4: Yep, you're right. It's wrong. I will update that. Thanks!

Luni-4:
> > How can we consider definition and implementation? Should I count one if I declare and implement the same `foo` method in the same file or two?
> 
> I think we should only consider raw count of definitions.

I was thinking of this case:

```
int foo();

int foo() {
 return 0;
}
```

Running `rust-code-analysis` dump:

```
`- {translation_unit:214} from (1, 1) to (6, 1)
   |- {declaration:236} from (1, 1) to (1, 11) : int foo();
   |  |- {primitive_type:61} from (1, 1) to (1, 4) : int
   |  |- {function_declarator:254} from (1, 5) to (1, 10) : foo()
   |  |  |- {identifier:1} from (1, 5) to (1, 8) : foo
   |  |  `- {parameter_list:277} from (1, 8) to (1, 10) : ()
   |  |     |- {(:17} from (1, 8) to (1, 9) : (
   |  |     `- {):8} from (1, 9) to (1, 10) : )
   |  `- {;:39} from (1, 10) to (1, 11) : ;
   `- {function_definition:235} from (3, 1) to (5, 2)
      |- {primitive_type:61} from (3, 1) to (3, 4) : int
      |- {function_declarator:254} from (3, 5) to (3, 10) : foo()
      |  |- {identifier:1} from (3, 5) to (3, 8) : foo
      |  `- {parameter_list:277} from (3, 8) to (3, 10) : ()
      |     |- {(:17} from (3, 8) to (3, 9) : (
      |     `- {):8} from (3, 9) to (3, 10) : )
      `- {compound_statement:263} from (3, 11) to (5, 2)
         |- {{:43} from (3, 11) to (3, 12) : {
         |- {return_statement:287} from (4, 2) to (4, 11) : return 0;
         |  |- {return:75} from (4, 2) to (4, 8) : return
         |  |- {number_literal:95} from (4, 9) to (4, 10) : 0
         |  `- {;:39} from (4, 10) to (4, 11) : ;
         `- {}:44} from (5, 1) to (5, 2) : }
```

We have `declaration` and `function definition`. Which one should I count? If we count both of them we would have two which is wrong, otherwise one.

But if we define this code and we have decided to count `declaration` it would be 0.
```
int foo() {
 return 0;
}
```

We could count `function_declarator` to solve the case above, but the problems remains for the first case, we would have 2 which is wrong. 

This explanation is to explain why I've used Strings :)

calixteman:
I think we should just count definition (i.e. `function_definition`).

src/nom.rs
```diff
@@ -0,0 +1,255 @@
...
+
+#[cfg(test)]
+mod tests {
+    use std::path::PathBuf;
+
+    use super::*;
+
+    #[test]
+    fn test_nom_python() {
+        let sample =
```
calixteman, line 205: Too much redundant code in these tests.
You could probably add a function in tools.rs to get a parser from a string and a filename (the language would be guessed with the file extension).

src/nom.rs
```diff
@@ -0,0 +1,262 @@
...
+#[cfg(test)]
+mod tests {
+    use std::path::PathBuf;
+
+    use super::*;
+
+    #[test]
+    fn test_nom_python() {
+        check_metrics!(
+            "def a():\n    pass\n\ndef b():\n    pass\n\ndef c():\n    pass\n",
```
calixteman, line 206: nit: add a lambda in the test.

src/macros.rs
```diff
@@ -161,3 +161,19 @@ macro_rules! color {
...
+
+#[macro_export]
+macro_rules! check_metrics {
+    ($source: expr, $file: expr, $parser: ident, $metric: ident,
+     [ $( ( $func: ident, $true_value: expr $(,$type: ty)? )$(,)* )* ]) => {
+        {
+            let sample = String::from($source);
+            let path = PathBuf::from($file);
+            let b_sample = sample.as_bytes().to_vec();
+            let parser = $parser::new(b_sample.clone(), &path, None);
```
calixteman, line 173: nit: replace `b_sample.clone()` by `$source.to_string().into_bytes()`

calixteman:
Thx

PR #140, opened 2020-04-03
Test rust-code-analysis on stable

Luni-4:
Test `rust-code-analysis` on `Linux` using the Rust stable version

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -41,7 +41,6 @@ tasks:
             - "-cx"
             - "curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3 get-pip.py &&
                rustup toolchain install stable &&
-               rustup toolchain install nightly &&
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -41,7 +41,6 @@ tasks:
             - "-cx"
             - "curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3 get-pip.py &&
                rustup toolchain install stable &&
-               rustup toolchain install nightly &&
                rustup default stable &&
```
marco-c, line 44: I wonder if this line is still needed.

Luni-4: Since tests are passing it's not needed anymore :)

marco-c: Well, the line is still there :P

Luni-4: Ouch, you're right!

PR #146, opened 2020-04-07
Fix CLOC metric

Luni-4:
- Fix wrong implementation for `CLOC`
- Add unit tests to verify the new implementation
- Add `CLOC` explanation to the README

### DIFF
--- a/src/loc.rs
+++ b/src/loc.rs
@@ -235,3 +235,48 @@ impl Loc for JavaCode {}
 impl Loc for GoCode {}
 impl Loc for CssCode {}
 impl Loc for HtmlCode {}
+
+#[cfg(test)]
+mod tests {
+    use std::path::PathBuf;
+
+    use super::*;
+
+    #[test]
+    fn test_cloc_python() {
+        // FIXME Add block comments
+        check_metrics!(
+            "# Line Comment\na = 42 # Line Comment\n",
+            "foo.py",
+            PythonParser,
+            loc,
+            [(cloc, 2, usize)]
+        );
+    }
+
+    #[test]
+    fn test_cloc_rust() {
+        check_metrics!(
+            "/*Block comment\nBlock Comment*/\n//Line Comment\n/*Block Comment*/ let a = 42; // Line Comment\n",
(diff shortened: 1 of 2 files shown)


src/loc.rs
```diff
@@ -92,29 +89,28 @@ where
...
         use Python::*;
 
-        let start = init(node, stats, is_func_space, is_unit);
+        let (start, end) = init(node, stats, is_func_space, is_unit);
 
         match node.kind_id().into() {
             String | DQUOTE | DQUOTE2 | ExpressionStatement | Block | Module => {}
             Comment => {
-                stats.comment_lines.insert(start);
+                stats.comment_lines += (end - start) + 1;
```
calixteman, line 113: you probably need to handle the case of multiline comments:
https://twitter.com/gvanrossum/status/112670605505077248

Luni-4: I added a `TODO` in the `Python` test because I don't know how to discriminate a string used as comment from a simple string

calixteman: An idea could be to look for string where the parent is an ExpressionStatement. I'd say that in the case the string is the statement and so this the only thing we've so a comment. Wdyt ?

Luni-4: Yep, good idea! I'm going to implement that, thanks @calixteman! :)

src/loc.rs
```diff
@@ -239,3 +241,48 @@ impl Loc for JavaCode {}
...
+
+    #[test]
+    fn test_cloc_python() {
+        check_metrics!(
+            "\"\"\"Block comment\nBlock comment\n\"\"\"\n
+            # Line Comment\na = 42 # Line Comment\n",
+            "foo.py",
+            PythonParser,
+            loc,
+            [(cloc, 5, usize)]
```
calixteman, line 259: mmh... I see only 4 comments here, am I correct ?

Luni-4:
```
"""Block comment
Block comment
"""
#Line Comment
a = 42 # Line Comment
```

There are 5 lines because I put a `\n` after the second `Block Comment`

PR #148, opened 2020-04-08
Fix clippy warnings

Luni-4:


skydancer0919:
hello,
i want just inputmapper files. all of them will you send my mail address?

Luni-4 <notifications@github.com>, 8 Nis 2020 Çar, 12:18 tarihinde şunu
yazdı:

src/web/server.rs
```diff
@@ -588,7 +588,9 @@ mod tests {
     #[actix_rt::test]
     async fn test_web_comment_plain_bad_chars() {
         let bad_bytes = &[142, 137, 138, 136, 140, 141, 10];
+        #[allow(clippy::string_lit_as_bytes)]
```
calixteman, line 591: nit: b"..." should be ok + adding bad_bytes: &[u8]

src/tools.rs
```diff
@@ -68,10 +69,10 @@ fn mode_to_str(mode: &[u8]) -> Option<String> {
     std::str::from_utf8(mode)
         .ok()
         .map(|m| m.to_lowercase())
-        .map(|m| m.to_string())
+        .map(|m| m)
```
calixteman, line 72: the lambda is identity so could be removed

src/web/server.rs
Luni-4: I can `concat` the two arrays because they have different size. How should I address that?

calixteman: I just tryed:
```rust
let bad_bytes: &[u8] = &[142, 137, 138, 136, 140, 141, 10];
let input_vec = [b"/*char*/s: ", bad_bytes].concat();
```
and it's ok

Luni-4: Oh, I forgot the & :sweat_smile:

Luni-4:
@calixteman 

Can we merge this one?

Luni-4:
Rebased!

PR #149, opened 2020-04-08
Remove clone trait for Config struct

Luni-4:
This PR removes the `Clone` trait for the `Config` struct

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -17,8 +17,7 @@ use walkdir::{DirEntry, WalkDir};
 use rust_code_analysis::web::server;
 use rust_code_analysis::*;
 
-// TODO: we could probably avoid to have to clone the Config...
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -39,11 +38,10 @@ struct Config {
 struct JobItem {
     language: Option<LANG>,
     path: PathBuf,
-    cfg: Config,
+    cfg: Arc<Config>,
 }
 
 type JobReceiver = Receiver<Option<JobItem>>;
```
calixteman, line 44: I don't really see the point to remove JobSender alias but ok: it isn't super used.
For consistency, you should remove JobReceiver too.

Luni-4: My bad. The code is clearer having it, thanks for spotting that! Fixed!

PR #152, opened 2020-04-15
Dump json on stdout

### LINKED ISSUE #106: Dump json to stdout
It would be helpful to dump json function spaces to `stdout`

Luni-4:
This PR fixes #106 

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -27,6 +27,7 @@ struct Config {
     count_filter: Vec<String>,
     function: bool,
     metrics: bool,
+    json: bool,
     output: String,
     pretty: bool,
     line_start: Option<usize>,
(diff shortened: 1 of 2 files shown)


calixteman:
I think it'd be better to have an option `-t, --type` with value equal to `json`, '`tabular`, whatever... to specify the type of output we want.
And then if at some point we've something else than metrics to output we can then reuse that stuff.

Luni-4:
> I think it'd be better to have an option `-t, --type` with value equal to `json`, '`tabular`, whatever... to specify the type of output we want.
> And then if at some point we've something else than metrics to output we can then reuse that stuff.

Yep, you're perfectly right. I'm still working on that, so we can close this one for now! :)

PR #157, opened 2020-04-22
Move enum command line tool in its own crate

Luni-4:


sylvestre:
Why? It seems overkill for now
Moving into a crate makes software pretty hard to package in linux distro.

Luni-4:
My original goal was to reduce the compilation time, moving the command line tool in its own crate allows the user to compile only the library.

`cargo build` builds the library
`cargo build --all` builds the command line tool also

Is it possible to get this outcome without damaging the packaging @sylvestre?

sylvestre:
it is a 66 lines of code file. I don't think compilation is a big deal here and rust handles incremental compilation.
I think you should keep the library simple and not over engineer it ;)

Luni-4:
> it is a 66 lines of code file. I don't think compilation is a big deal here and rust handles incremental compilation.
> I think you should keep the library simple and not over engineer it ;)

I agree with you, but I don't think I've over engineered a lot here, we can get the same results as before just adding the `--all` option to the `cargo build` command. At the same time we allow the user, which are not interested in the command line tool, to build only the library.

Anyway, it's not a problem for me if the code remains like it is now, so feel free to close this PR anytime :wink:

PR #171, opened 2020-05-11
Replace env variables with let variables

Luni-4:


.taskcluster.yml
```diff
@@ -236,6 +236,8 @@ tasks:
         then:
           $let:
             version: {$eval: 'head_branch[10:]'}
+            linux_artifact: rust-code-analysis-linux-x86_64.tar.gz
+            windows_artifact: rust-code-analysis-windows-x86_64.zip
```
marco-c, line 240: Since you're here, let's also call this artifact `rust-code-analysis-win-x86_64.zip` (this naming is used more commonly). You'll need to update the `build_windows_release` task too.

marco-c:
> We could

I'm not sure what I meant to write here :P

PR #190, opened 2020-05-20
Add a Matrix badge

Luni-4:


marco-c:
I like it, we should add it on our other projects!

Luni-4:
Yep, it's nice! :)

PR #191, opened 2020-05-21
Add a node variant

Luni-4:
This variant hides the crate used to handle nodes in the code.

### DIFF
--- a/src/alterator.rs
+++ b/src/alterator.rs
@@ -1,5 +1,3 @@
-use tree_sitter::Node;
-
 use crate::*;
 
 pub trait Alterator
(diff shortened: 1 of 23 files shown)


calixteman:
Is this patch just aim to hide the `use tree_sitter::Node;` ?
Could you explain the exact rational ?

Luni-4:
> Is this patch just aim to hide the `use tree_sitter::Node;` ?
> Could you explain the exact rational ?

Yep, the main goal is hiding `tree_sitter::Node` to users since in the documentation the `Node` struct is exported and I can't avoid showing it because that would mean a lack of documentation.

That's why I wanted to hide the library that handles nodes, at the same time the entire `tree_sitter` library is completely handled internally by `rust-code-analysis`

Luni-4:
This one seems a bit complicated instead, I couldn't think an alternative solution

calixteman:
Please rebase.

Luni-4:
Done.

PR #206, opened 2020-06-03
Make preprocess function thread-independent

Luni-4:
This PR makes the `preprocess` function thread-independent.


Luni-4:
Rebased and ready to be merged.

PR #209, opened 2020-06-11
Print supported langauges

### LINKED ISSUE #159: Print supported languages in the CLI tool
Print all languages supported by `rust-code-analysis` in the cli tool

Luni-4:
This PR fixes #159 

### DIFF
--- a/rust-code-analysis-cli/src/main.rs
+++ b/rust-code-analysis-cli/src/main.rs
@@ -248,6 +248,12 @@ fn main() {
         .author(&*env!("CARGO_PKG_AUTHORS").replace(':', "\n"))
         .about("Analyze source code")
         .arg(
+            Arg::with_name("show_languages")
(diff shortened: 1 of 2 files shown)


calixteman:
Can you change option name to `languages` and only have a long one (I mean only `--languages`) ?
The 's' could be useful for something else and it's a pity to waste a char just for an option you almost never use (it's just an 'out of curiosity' option).

Luni-4:
> Can you change option name to `languages` and only have a long one (I mean only `--languages`) ?
> The 's' could be useful for something else and it's a pity to waste a char just for an option you almost never use (it's just an 'out of curiosity' option).

Yeah, this is a strange command indeed, because we now have the documentation. Perhaps we should close this PR and the relative issue. It was required when there wasn't any documentation, but I think it is now unnecessary. What do you think?

calixteman:
Agreed... useless

PR #211, opened 2020-06-11
Deploy documentation automatically

Luni-4:
Once we release, the documentation is deployed automatically.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -291,9 +291,9 @@ tasks:
               command:
                 - "/bin/sh"
                 - "-cx"
-                - "git config --global user.email you@example.com &&
+                - "git config --global user.email moz-tools-bot@moz.tools &&
                    git config --global user.name moz.tools Bot &&
-                   git clone --quiet ${repository} &&
+                   git clone --recursive --quiet ${repository} &&
                    cd rust-code-analysis &&
                    taskboot retrieve-artifact --output-path=. \
                             --artifacts=public/book.tar.gz &&
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -291,13 +291,13 @@ tasks:
               command:
                 - "/bin/sh"
                 - "-cx"
-                - "git clone --recursive --quiet https://github.com/SoftengPoliTo/rust-code-analysis &&
+                - "git config --global user.email you@example.com &&
+                   git config --global user.name moz.tools Bot &&
+                   git clone --quiet ${repository} &&
                    cd rust-code-analysis &&
                    taskboot retrieve-artifact --output-path=. \
                             --artifacts=public/book.tar.gz &&
                    tar xfz book.tar.gz -C rust-code-analysis-book &&
-                   git config --global user.email you@example.com &&
-                   git config --global user.name Your Name &&
                    ./rust-code-analysis-book/deploy-to-GitHub-Pages &&
                    taskboot git-push --force-push \
                             github.com/mozilla/rust-code-analysis \
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -56,6 +56,36 @@ tasks:
           owner: cdenizet@mozilla.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+      - taskId: {$eval: as_slugid("build_documentation")}
+        created: {$fromNow: ''}
+        deadline: {$fromNow: '2 hour'}
+        provisionerId: proj-relman
+        workerType: ci
+        payload:
+          maxRunTime: 3600
+          image: "rust:buster"
+          command:
+            - "/bin/bash"
+            - "-cx"
+            - "git clone --recursive --quiet ${repository} &&
+               cd rust-code-analysis &&
+               git -c advice.detachedHead=false checkout ${head_rev} &&
+               cargo install mdbook --no-default-features --features search,output --vers \"^0.1.0\" &&
+               cargo doc --release &&
+               cd rust-code-analysis-book &&
+               mdbook build &&
+               tar -zvcf /book.tar.gz book/"
+          artifacts:
+            public/book.tar.gz:
+              expires: {$fromNow: '2 weeks'}
+              path: /book.tar.gz
+              type: file
+        metadata:
+          name: rust-code-analysis documentation
+          description: rust-code-analysis build documentation
+          owner: cdenizet@mozilla.com
+          source: ${repository}/raw/${head_rev}/.taskcluster.yml
+
       - taskId: {$eval: as_slugid("ccov_test_task")}
         created: {$fromNow: ''}
         deadline: {$fromNow: '1 hour'}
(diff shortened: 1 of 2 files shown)


Luni-4:
I agree on the documentation task (for 'cargo doc` and the `mdbook` command), but I think we should avoid creating an artifact for the release task. The artifact could be too big in terms of size. So the best thing to do is the commit above in my opinion.

marco-c:
> I agree on the documentation task (for cargo doc and the mdbook command), but I think we should avoid creating an artifact for the release task. The artifact could be too big in terms of size. So the best thing to do is the commit above in my opinion.

We should check how big it is, I would expect it to be relatively small given that the documentation is mostly text.

Luni-4:
> > I agree on the documentation task (for cargo doc and the mdbook command), but I think we should avoid creating an artifact for the release task. The artifact could be too big in terms of size. So the best thing to do is the commit above in my opinion.
> 
> We should check how big it is, I would expect it to be relatively small given that the documentation is mostly text.

Ok, checked the size: 1,7MB. But why can't we automatically run the script in the release task? After all, we are uploading the output of `mdbook` as Github Pages. So if `mdbook` and `cargo doc` commands are already tested in the documentation task, perhaps it's a boilerplate loading the artifact and run the script again

marco-c:
> Ok, checked the size: 1,7MB. But why can't we run the script in the release task? After all, we are uploading the output of mdbook as Github Pages. Since mdbook and cargo doc commands will be already tested in the documentation task, perhaps it's a boilerplate loading the artifact and run the script again

It's better to have logical separation between the tasks: one task does the build (in this case documentation build), another task (very lightweight) deploys it (in this case on gh-pages).
If you run the build command on the release task, you'll make it slower (as you need to rebuild) and heavier (as you need the dependencies to build).

.taskcluster.yml
```diff
@@ -56,6 +56,59 @@ tasks:
...
+              expires: {$fromNow: '2 weeks'}
+              path: rust-code-analysis-book/book
+              type: file
+        metadata:
+          name: rust-code-analysis documentation
+          description: rust-code-analysis build documentation
+          owner: cdenizet@mozilla.com
+          source: ${repository}/raw/${head_rev}/.taskcluster.yml
+
+      - taskId: {$eval: as_slugid("test_gh_pages")}
```
marco-c, line 88: You can remove this task and put its commands in the `deploy_release` task

Luni-4: Yep, just testing for now. If it works I'm going to remove that task

Luni-4:
@marco-c 
This PR is now ready

.taskcluster.yml
```diff
@@ -56,6 +56,36 @@ tasks:
...
+        payload:
+          maxRunTime: 3600
+          image: "rust:buster"
+          command:
+            - "/bin/bash"
+            - "-cx"
+            - "git clone --recursive --quiet ${repository} &&
+               cd rust-code-analysis &&
+               git -c advice.detachedHead=false checkout ${head_rev} &&
+               cargo install mdbook --no-default-features --features search,output --vers \"^0.1.0\" &&
```
marco-c, line 73: What are the default features?

marco-c, line 73: Why `--vers \"^0.1.0\"`?

.taskcluster.yml
```diff
@@ -56,6 +56,36 @@ tasks:
...
+        deadline: {$fromNow: '2 hour'}
+        provisionerId: proj-relman
+        workerType: ci
+        payload:
+          maxRunTime: 3600
+          image: "rust:buster"
+          command:
+            - "/bin/bash"
+            - "-cx"
+            - "git clone --recursive --quiet ${repository} &&
```
marco-c, line 70: I wonder if we need a recursive clone for the docs

.taskcluster.yml
```diff
@@ -254,13 +285,25 @@ tasks:
...
               maxRunTime: 3600
-              image: "mozilla/taskboot:0.2.5"
+              image: "mozilla/taskboot:0.2.7"
               env:
                 TASKCLUSTER_SECRET: project/relman/rust-code-analysis/deploy
               command:
                 - "/bin/sh"
                 - "-cx"
-                - "taskboot github-release mozilla/rust-code-analysis ${version} --asset ${linux_artifact}:public/${linux_artifact} ${windows_artifact}:public/${windows_artifact}"
+                - "git clone --recursive --quiet https://github.com/SoftengPoliTo/rust-code-analysis &&
```
marco-c, line 294: Wrong repo :)

.taskcluster.yml
```diff
@@ -254,13 +285,25 @@ tasks:
...
+                   cd rust-code-analysis &&
+                   taskboot retrieve-artifact --output-path=. \
+                            --artifacts=public/book.tar.gz &&
+                   tar xfz book.tar.gz -C rust-code-analysis-book &&
+                   git config --global user.email you@example.com &&
```
marco-c, line 299: Nit: move the `git config` commands right after the `cd`, so it's clear it's related.

.taskcluster.yml
```diff
@@ -254,13 +285,25 @@ tasks:
...
+                   git config --global user.name Your Name &&
```
marco-c, line 300: Use `moz.tools Bot` instead of `Your Name`

.taskcluster.yml
Luni-4: I just followed the `mdbook' [README](https://github.com/rust-lang/mdBook#what-does-it-look-like) section called `Note for automatic deployment`

Luni-4: Same as above

.taskcluster.yml
Luni-4: Not for now, we can remove that option

marco-c: Sorry I didn't get your answer, can we or can we not remove the option?

.taskcluster.yml
marco-c: :+1:

.taskcluster.yml
```diff
@@ -254,13 +285,25 @@ tasks:
...
               maxRunTime: 3600
-              image: "mozilla/taskboot:0.2.5"
+              image: "mozilla/taskboot:0.2.7"
               env:
                 TASKCLUSTER_SECRET: project/relman/rust-code-analysis/deploy
               command:
                 - "/bin/sh"
                 - "-cx"
-                - "taskboot github-release mozilla/rust-code-analysis ${version} --asset ${linux_artifact}:public/${linux_artifact} ${windows_artifact}:public/${windows_artifact}"
+                - "git config --global user.email you@example.com &&
```
marco-c, line 294: Nit: use `moz-tools-bot@moz.tools`, even if it's fake, it's nicer than example.com :P

.taskcluster.yml
Luni-4: For now it's not necessary, but I guess that if we will add the `cargo publish` options it could be necessary for the documentation of the `tree-sitter` dependencies. Perhaps we could maintain it in order to be sure. Wdyt?

marco-c: OK, let's keep it, it doesn't hurt that much.

PR #217, opened 2020-06-22
Publish rust-code-analysis automatically on crates.io

Luni-4:
This PR publishes `rust-code-analysis` automatically.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -303,7 +303,8 @@ tasks:
                             github.com/mozilla/rust-code-analysis \
                             moz-tools-bot \
                             gh-pages &&
-                   taskboot github-release mozilla/rust-code-analysis ${version} --asset ${linux_artifact}:public/${linux_artifact} ${windows_artifact}:public/${windows_artifact}"
+                   taskboot github-release mozilla/rust-code-analysis ${version} --asset ${linux_artifact}:public/${linux_artifact} ${windows_artifact}:public/${windows_artifact} &&
+                   taskboot cargo-publish"
             metadata:
               name: "rust-code-analysis release publication ${version}"
               description: rust-code-analysis release publication on Github


.taskcluster.yml
```diff
@@ -241,14 +241,14 @@ tasks:
...
-            image: "mozilla/taskboot:0.2.5"
+            image: "mozilla/taskboot:0.2.7"
             env:
               WINDOWS_BIN: rust-code-analysis-cli.exe
             command:
                 - "/bin/sh"
                 - "-cx"
                 - "taskboot retrieve-artifact --output-path=. \
-                            --artifacts=$WINDOWS_BIN:public/$WINDOWS_BIN &&
+                            --artifacts=public/$WINDOWS_BIN &&
```
marco-c, line 251: I don't remember, did we change the `retrieve-artifact` command recently?

Anyway, this change seems unrelated to this PR, so please open a separate PR with it.

Luni-4: Yep, we [did](https://github.com/mozilla/task-boot/commit/34c6274fd1af09b0bce163ec4649ab0deca245e4). 
I thought I could use the same PR to update everything :P
But you're right, I'm going to open a new PR

marco-c: Better to have each PR be a different set of logic changes (or each commit in a PR, but that's harder to handle as sometimes you'd have to rewrite history).

Luni-4: I agree, but since they were both related to `TaskCluster`,I thought I could have used two different commits

marco-c:
It's a pity that now the `deploy_release` task will have to perform a build, but that's OK.

Luni-4:
> It's a pity that now the `deploy_release` task will have to perform a build, but that's OK.

I know, but the entire process happens once in a while

PR #218, opened 2020-06-23
Add cognitive complexity

Luni-4:
This PR adds the `Cognitive Complexity` metric. It has been used this [document](https://www.sonarsource.com/docs/CognitiveComplexity.pdf) to implement it.


src/asttools.rs
```diff
@@ -54,3 +55,44 @@ macro_rules! has_ancestors {
...
+
+#[macro_use]
+macro_rules! number_of_ancestors {
+    ($node:expr, $( $typs:pat )|*, $( $stops:pat )|*) => {{
+        let mut count = 0;
+        let mut node = *$node;
+        while let Some(parent) = node.parent() {
+            match parent.kind_id().into() {
+                $( $typs )|* => {
+                    if !Self::is_else_if(&parent) {
```
calixteman, line 87: I think the macro name doesn't reflect what it's doing

src/checker.rs
```diff
@@ -82,6 +87,14 @@ impl Checker for CppCode {
         let code = &code[node.start_byte()..node.end_byte()];
         AC.is_match(code)
     }
+
+    #[inline(always)]
+    fn is_else_if(node: &Node) -> bool {
+        if let Some(parent) = node.parent() {
+            return node.kind_id() == Cpp::IfStatement && parent.kind_id() == Cpp::IfStatement;
```
calixteman, line 94: no need to get the parent if node.kind != IfStatement

src/checker.rs
```diff
@@ -296,6 +309,14 @@ impl Checker for RustCode {
         code.starts_with(b"/// cbindgen:")
     }
 
+    #[inline(always)]
+    fn is_else_if(node: &Node) -> bool {
+        if let Some(parent) = node.parent() {
+            return node.kind_id() == Rust::IfExpression && parent.kind_id() == Rust::IfExpression;
```
calixteman, line 315: same as above

src/metrics/cognitive.rs
```diff
@@ -0,0 +1,808 @@
...
+use serde::Serialize;
+use std::fmt;
+use tree_sitter::Node;
+
+use crate::checker::Checker;
+use crate::*;
+
+// TODO: Find a way to increment the cognitive complexity value
+// for recursive code.
+// The method SHOULD work for EACH of the supported languages.
```
calixteman, line 11: You should explain why it isn't trivial (depending of the language)

src/metrics/cognitive.rs
```diff
@@ -0,0 +1,808 @@
...
+            if let $( $typs )|* = child.kind_id().into() {
+                $stats.structural = $stats
+                    .boolean_seq
+                    .eval_based_on_prev(child.kind_id(), $stats.structural);
+            }
+        }
+    };
+}
+
+macro_rules! nesting_levels {
```
calixteman, line 81: I wonder if we could use a stack in spaces.rs like the one we've for spaces themselves to avoid to walk too much in the tree.

src/metrics/cognitive.rs
```diff
@@ -0,0 +1,808 @@
...
+
+    fn not_operator(&mut self, not_id: u16) {
+        self.first_boolean = self.boolean_op.is_none();
+        self.boolean_op = Some(not_id);
+    }
+
+    fn eval_based_on_prev(&mut self, bool_id: u16, structural: usize) -> usize {
+        let mut new_structural = structural;
+        if let Some(prev) = self.boolean_op {
+            new_structural = if prev != bool_id && !self.first_boolean {
```
calixteman, line 153: could you add a comment to explain each of the different cases ?

src/metrics/cognitive.rs
```diff
@@ -0,0 +1,808 @@
...
+            }
+            BinaryExpression2 => {
+                compute_booleans!(node, stats, AMPAMP | PIPEPIPE);
+            }
+            _ => {}
+        }
+    }
+}
+
+impl Cognitive for MozjsCode {}
```
calixteman, line 321: It'd be nice to support js too since we've this language in mozilla-central.

src/metrics/cognitive.rs
```diff
@@ -0,0 +1,808 @@
...
+                 return bar(foo(a))(a)\n",
+            "foo.py",
+            PythonParser,
+            cognitive,
+            [(cognitive, 5, usize)]
+        );
+    }
+
+    #[test]
+    fn test_real_function_cognitive() {
```
calixteman, line 786: A lot of tests... :+1:

src/metrics/cognitive.rs
Luni-4: I propose to iterate over this in another PR. What do you think?

src/checker.rs
```diff
@@ -82,6 +87,17 @@ impl Checker for CppCode {
         let code = &code[node.start_byte()..node.end_byte()];
         AC.is_match(code)
     }
+
+    #[inline(always)]
+    fn is_else_if(node: &Node) -> bool {
```
calixteman, line 92: The three next functions look pretty similar, probably we could use a macro which will take the node kind as argument.

src/metrics/cognitive.rs
```diff
@@ -0,0 +1,1008 @@
...
+            }
+            BinaryExpression => {
+                compute_booleans!(node, stats, AMPAMP | PIPEPIPE);
+            }
+            _ => {}
+        }
+    }
+}
+
+impl Cognitive for JavascriptCode {}
```
calixteman, line 402: you could add it for js and ts (should the same code as mozjs)

calixteman (approved):
Looks pretty good. Thx a lot for this patch.

PR #221, opened 2020-06-24
Make public the function to dump metrics

Luni-4:

### DIFF
--- a/src/output/dump_metrics.rs
+++ b/src/output/dump_metrics.rs
@@ -11,7 +11,37 @@ use crate::nom;
 
 use crate::spaces::{CodeMetrics, FuncSpace};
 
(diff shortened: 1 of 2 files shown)


src/output/dump_metrics.rs
```diff
@@ -11,7 +11,37 @@ use crate::nom;
...
+///
+/// # fn main() {
+/// let source_code = "int a = 42;";
+///
+/// // The path to a dummy file used to contain the source code
+/// let path = PathBuf::from("foo.c");
+/// let source_as_vec = source_code.as_bytes().to_vec();
+///
+/// // The parser of the code, in this case a CPP parser
+/// let parser = CppParser::new(source_as_vec.clone(), &path, None);
```
calixteman, line 33: no need to clone

PR #224, opened 2020-06-25
Change Halstead implementation

Luni-4:
In this way `CodeMetrics` is independent from lifetimes.

### DIFF
--- a/src/metrics/halstead.rs
+++ b/src/metrics/halstead.rs
@@ -9,13 +9,57 @@ use crate::getter::Getter;
 use crate::*;
 
 /// The `Halstead` metric suite.
-#[derive(Default, Debug)]
-pub struct Stats<'a> {
+#[derive(Default, Clone, Debug)]
+pub struct Stats {
+    u_operators: u64,
+    operators: u64,
+    u_operands: u64,
+    operands: u64,
+}
+
+/// Specifies the type of nodes accepted by the `Halstead` metric.
+pub enum HalsteadType {
+    /// The node is an `Halstead` operator
+    Operator,
+    /// The node is an `Halstead` operand
+    Operand,
+    /// The node is unknown to the `Halstead` metric
+    Unknown,
+}
+
+#[doc(hidden)]
+#[derive(Debug, Default, Clone)]
+pub struct HalsteadMaps<'a> {
     operators: FxHashMap<u16, u64>,
     operands: FxHashMap<&'a [u8], u64>,
 }
 
-impl Serialize for Stats<'_> {
+impl<'a> HalsteadMaps<'a> {
+    pub(crate) fn new() -> Self {
+        HalsteadMaps {
+            operators: FxHashMap::default(),
+            operands: FxHashMap::default(),
+        }
+    }
+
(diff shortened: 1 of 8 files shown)


calixteman:
Could you make a PR with the tests only and then we can see if the refactoring doesn't introduce any regression ?

Luni-4:
> Could you make a PR with the tests only and then we can see if the refactoring doesn't introduce any regression ?

Yep, good idea, thanks! :)

src/metrics/halstead.rs
```diff
@@ -76,39 +113,37 @@ impl<'a> fmt::Display for Stats<'a> {
...
+impl Stats {
     /// Merges a second `Halstead` metric suite into the first one
-    pub fn merge(&mut self, other: &Stats<'a>) {
-        for (k, v) in other.operators.iter() {
-            *self.operators.entry(*k).or_insert(0) += v;
-        }
-        for (k, v) in other.operands.iter() {
-            *self.operands.entry(*k).or_insert(0) += v;
-        }
+    pub fn merge(&mut self, other: &Stats) {
```
calixteman, line 118: we shouldn't have this function.
You can't sum the u_***: the maps need to be merged and then the new values are the length of the new maps.

src/metrics/halstead.rs
```diff
@@ -194,53 +228,104 @@ fn get_id<'a>(node: &Node<'a>, code: &'a [u8]) -> &'a [u8] {
...
+    code: &'a [u8],
+    stats: &mut Stats,
+    halstead_maps: &mut HalsteadMaps<'a>,
+) {
     match T::get_op_type(&node) {
-        HalsteadType::Operator => *stats.operators.entry(node.kind_id()).or_insert(0) += 1,
-        HalsteadType::Operand => *stats.operands.entry(get_id(node, code)).or_insert(0) += 1,
+        HalsteadType::Operator => {
+            *halstead_maps.operators.entry(node.kind_id()).or_insert(0) += 1;
+            stats.u_operators = halstead_maps.operators.len() as u64;
```
calixteman, line 240: I think it's useless to update this value each time we insert a new operators.
Maybe we can add a `finalize` function called when we just finish parsing to set u_****.

src/metrics/halstead.rs
```diff
@@ -194,53 +228,104 @@ fn get_id<'a>(node: &Node<'a>, code: &'a [u8]) -> &'a [u8] {
...
+            stats.operators += 1;
+        }
+        HalsteadType::Operand => {
+            *halstead_maps
+                .operands
+                .entry(get_id(node, code))
+                .or_insert(0) += 1;
+            stats.u_operands = halstead_maps.operands.len() as u64;
```
calixteman, line 248: see above

calixteman:
Please rebase.

Luni-4:
The new `finalize` approach breaks the tests, it seems the `Halstead` is no more computed in the right way.

Luni-4:
Now it works, I forgot to update the `Mi` implementation and update stacks :P

Luni-4:
@calixteman 

Can we merge this PR, please?

PR #225, opened 2020-06-25
Add Halstead tests

Luni-4:
This PR adds tests for the `Halstead` metric.


calixteman:
In the python code, could you add some inner functions:
```python
def foo:
  def bar:
    def toto:...
```
with some common variables and not common ones to check that the merge stuff is ok too ? (things like a+b... to have some operands and operators)

PR #231, opened 2020-07-01
Refactor loc tests

Luni-4:

### DIFF
--- a/src/metrics/loc.rs
+++ b/src/metrics/loc.rs
@@ -379,7 +379,9 @@ mod tests {
     #[test]
     fn test_blank() {
         check_metrics!(
-            "\na = 42\n\n",
+            "
+            a = 42
+            \n",
             "foo.py",
             PythonParser,
             loc,
@@ -387,7 +389,9 @@ mod tests {
         );
 
         check_metrics!(
-            "\nlet a = 42;\n\n",
+            "
+            let a = 42;
+            \n",
             "foo.rs",
             RustParser,
             loc,
@@ -403,7 +407,9 @@ mod tests {
         );
 
         check_metrics!(
-            "\nint a = 42;\n\n",
+            "
+            int a = 42;
+            \n",
             "foo.c",
             CppParser,
             loc,
@@ -414,8 +420,11 @@ mod tests {
     #[test]
     fn test_cloc() {
         check_metrics!(
-            "\"\"\"Block comment\nBlock comment\n\"\"\"\n
-            # Line Comment\na = 42 # Line Comment\n",
+            "\"\"\"Block comment
+            Block comment
+            \"\"\"
+            # Line Comment
+            a = 42 # Line Comment\n",
             "foo.py",
             PythonParser,
             loc,
@@ -423,7 +432,10 @@ mod tests {
         );
 
         check_metrics!(
-            "/*Block comment\nBlock Comment*/\n//Line Comment\n/*Block Comment*/ let a = 42; // Line Comment\n",
+            "/*Block comment
+             Block Comment*/
+             //Line Comment
+             /*Block Comment*/ let a = 42; // Line Comment\n",
             "foo.rs",
             RustParser,
             loc,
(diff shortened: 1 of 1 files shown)


src/metrics/loc.rs
```diff
@@ -480,28 +480,47 @@ mod tests {
...
+                      b,
+                      c):
+                 print(a)
+                 print(b)
+                 print(c)\n",
+            "foo.py",
+            PythonParser,
+            loc,
+            [
+                (sloc, 6, usize),
```
calixteman, line 518: Please add a comment for each value to explain why the value (exactly as you did for Halstead).

src/metrics/loc.rs
```diff
@@ -377,20 +377,21 @@ mod tests {
     use super::*;
 
     #[test]
-    fn test_blank_python() {
+    fn test_blank() {
         check_metrics!(
-            "\na = 42\n\n",
+            "
+            a = 42
+            \n",
```
marco-c, line 384: Why are you removing `\n` sometimes and keeping it other times?
I'd prefer to be consistent and either always have `\n` or never have it.

src/metrics/loc.rs
```diff
@@ -377,20 +377,21 @@ mod tests {
...
             "foo.py",
             PythonParser,
             loc,
             [(blank, 1, usize)]
         );
-    }
 
-    #[test]
-    fn test_blank_rust() {
```
marco-c, line 391: Maybe better to keep the tests as separated, so they can pass/fail individually

src/metrics/loc.rs
```diff
@@ -480,28 +480,47 @@ mod tests {
...
                 break;\n",
             "foo.c",
             CppParser,
             loc,
             [(lloc, 2, usize)]
         );
     }
+
+    #[test]
+    fn test_general_loc() {
```
marco-c, line 506: Since this is a new test and is not part of the refactoring of the pre-existing tests, it should be part of a separate PR

src/metrics/loc.rs
Luni-4: It's better to group tests on the same stuff in a single function, at least we immediately know the wrong metric. Having the same metric test for each language is a boilerplate to maintain and eventually update

Luni-4:
Closing in favor of #239

src/metrics/loc.rs
marco-c: If we keep them separated, we still immediately know the wrong metric, but we have the possibility to know if a specific language is broken.

Luni-4: I can't figure out the problem, sorry :)
Since each sub-test of a metric test is a macro, if a metric test is failing, we can immediately detect which macro is involved and at the same time we can determine if that specific language is broken.

marco-c: If there is a failure on the first assertion, the other assertions won't be evaluated.

In this case, if parsing Python fails, we won't parse Rust and C too, so we won't be able to tell right away if it's just a Python problem or if is a cross-language problem.

Luni-4: > If there is a failure on the first assertion, the other assertions won't be evaluated.
> 
> In this case, if parsing Python fails, we won't parse Rust and C too, so we won't be able to tell right away if it's just a Python problem or if is a cross-language problem.

Oh, so it's just a comfort matter. I thought to solve that problem commenting the series of sub-test macros until the botched one is found. Because if a test is failing we should get the hands on the same to discover the cause.

Actually I refactored the code because I didn't like all those little tests for the same thing

Luni-4: Anyway, I see your point, but I think it would be clear to have a test-per-metric instead of different test-per-metric-per-language as I thought initially

marco-c: Well, since there is no practical advantage in sticking the tests together, but there is an advantage in keeping them separated, let's keep them separated.

Luni-4: I don't like this approach, but if it is more comfortable for you, let's keep them as I thought initially :P

PR #239, opened 2020-07-06
Do not consider the spurious blank lines of a file

### LINKED ISSUE #143: Sloc shouldn't count spurious blank lines after the code
It would be possible to create a file which has a series of blank lines after the last line of code.

Luni-4:
This PR fixes #143 

### DIFF
--- a/src/tools.rs
+++ b/src/tools.rs
@@ -26,6 +26,8 @@ pub fn read_file(path: &PathBuf) -> std::io::Result<Vec<u8>> {
     let mut data = Vec::new();
     file.read_to_end(&mut data)?;
 
+    remove_blank_lines(&mut data);
+
     Ok(data)
 }
 
@@ -80,13 +82,7 @@ pub fn read_file_with_eol(path: &PathBuf) -> std::io::Result<Option<Vec<u8>>> {
 
     file.read_to_end(&mut data)?;
 
-    if let Some(c) = data.last() {
-        if *c != b'\n' {
-            data.push(b'\n');
-        }
-    } else {
-        data.push(b'\n');
-    }
+    remove_blank_lines(&mut data);
 
     Ok(Some(data))
 }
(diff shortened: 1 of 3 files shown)


Luni-4:
@marco-c 

All tests here are coherent with the fix, they are necessary. So we can create a single commit for this one

marco-c:
> All tests here are coherent with the fix, they are necessary. So we can create a single commit for this one

Well, they are not all necessary for the fix. E.g. merging the tests into a single test is not necessary.
Please make self-contained PRs, so they are easier to review and easier to backout individually should they cause problems.

Luni-4:
> > All tests here are coherent with the fix, they are necessary. So we can create a single commit for this one
> 
> Well, they are not all necessary for the fix. E.g. merging the tests into a single test is not necessary.
> Please make self-contained PRs, so they are easier to review and easier to backout individually should they cause problems.

Done

src/tools.rs
```diff
@@ -243,6 +239,13 @@ pub fn guess_language<P: AsRef<Path>>(buf: &[u8], path: P) -> (Option<LANG>, Str
     }
 }
 
+pub(crate) fn remove_blank_lines(data: &mut Vec<u8>) {
+    while data.last().map_or_else(|| false, |c| *c == b'\n') {
```
calixteman, line 243: You could just count the trailing \n and then `resize` the vector correctly or push a new \n if no \n.

src/tools.rs
```diff
@@ -243,6 +239,15 @@ pub fn guess_language<P: AsRef<Path>>(buf: &[u8], path: P) -> (Option<LANG>, Str
     }
 }
 
+pub(crate) fn remove_blank_lines(data: &mut Vec<u8>) {
```
calixteman, line 242: I'd say that 99.9% of the time the file has at least one trailing EOL. So I think we should compute `count_trailing` and then if !0 then `truncate` (which is better than resize) else push an EOL.

calixteman (approved):

PR #260, opened 2020-08-06
Update Mi test

Luni-4:
Use the new cyclomatic implementation to test the Mi metric


Luni-4:
@calixteman 

This PR fixes the broken CI, so we can land it anytime

PR #270, opened 2020-09-14
Fix rust-code-analysis for mozilla-central

Luni-4:
- Fix a wrong logic
- Remove a blocking `unwrap` function

### DIFF
--- a/src/spaces.rs
+++ b/src/spaces.rs
@@ -180,23 +180,26 @@ fn compute_halstead_and_mi<'a, T: ParserTrait>(state: &mut State<'a>) {
 }
 
 fn finalize<'a, T: ParserTrait>(state_stack: &mut Vec<State<'a>>, diff_level: usize) {
+    if state_stack.is_empty() {
+        return;
+    }
     for _ in 0..diff_level {
-        if state_stack.len() <= 1 {
+        if state_stack.len() == 1 {
             let mut last_state = state_stack.last_mut().unwrap();
             compute_halstead_and_mi::<T>(&mut last_state);
             break;
+        } else {
+            let mut state = state_stack.pop().unwrap();
+            compute_halstead_and_mi::<T>(&mut state);
+
+            let mut last_state = state_stack.last_mut().unwrap();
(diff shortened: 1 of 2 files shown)


src/spaces.rs
```diff
@@ -176,22 +176,24 @@ fn compute_halstead_and_mi<'a, T: ParserTrait>(state: &mut State<'a>) {
 
 fn finalize<'a, T: ParserTrait>(state_stack: &mut Vec<State<'a>>, diff_level: usize) {
     for _ in 0..diff_level {
-        if state_stack.len() <= 1 {
+        if state_stack.len() < 1 {
```
calixteman, line 179: nit: check for stack emptyness before the loop

src/spaces.rs
```diff
@@ -176,22 +176,24 @@ fn compute_halstead_and_mi<'a, T: ParserTrait>(state: &mut State<'a>) {
...
+            break;
+        } else if state_stack.len() == 1 {
```
calixteman, line 181: the first line of this block and the next one are almost the same

calixteman (dismissed):
Looks good. Just fix the nits to have a nicer code.

src/spaces.rs
Luni-4: I can't move that `else if` outside the loop though. What is the problem here?

calixteman: I thought you did almost the same thing in the two branches but I think your code is correct because we need to keep the last value int he stack.

Luni-4: Oh, I see. I thought there was a problem with the loop, sorry :)

src/spaces.rs
```diff
@@ -180,23 +180,26 @@ fn compute_halstead_and_mi<'a, T: ParserTrait>(state: &mut State<'a>) {
 }
 
 fn finalize<'a, T: ParserTrait>(state_stack: &mut Vec<State<'a>>, diff_level: usize) {
+    if state_stack.len() < 1 {
```
calixteman, line 183: nit: is_empty ?

calixteman (approved):
LGTM, thx

PR #276, opened 2020-09-24
Fix wrong Halstead computation for Python

Luni-4:
We were counting wrong operators for Python Halstead's implementation.

### DIFF
--- a/src/getter.rs
+++ b/src/getter.rs
@@ -45,14 +45,15 @@ impl Getter for PythonCode {
 
         let id = node.object().kind_id();
         match id.into() {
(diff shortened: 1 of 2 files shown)


sylvestre:
You should add a test for this check :)

Luni-4:
> You should add a test for this check :)

It's already tested [here](https://github.com/mozilla/rust-code-analysis/blob/master/src/metrics/halstead.rs#L306) :)

Luni-4:
@sylvestre 

Added a regression test also :)

calixteman (approved):

PR #277, opened 2020-09-25
Add a way to retrieve and print operators and operands

### LINKED ISSUE #232: Dump Halstead operators and operands
it would be helpful to add a flag to dump all `Halstead` operators and operands present in a code.

Luni-4:
This PR fixes #232


Luni-4:
This PR needs to be rebased, simplified a bit and some tests should be added

src/languages/language_ccomment.rs
```diff
@@ -21,38 +21,41 @@ pub enum Ccomment {
...
-            Ccomment::PreprocContinuationLine,
-        ),
+        ("Nothing", "nothing"),
+        ("RawStringLiteral", "raw_string_literal"),
+        ("StringLiteralToken1", "string_literal_token1"),
+        ("DefineToken1", "define_token1"),
+        ("CharLiteralToken1", "char_literal_token1"),
+        ("CharLiteral", "char_literal"),
+        ("Define", "define"),
+        ("TranslationUnitRepeat1", "translation_unit_repeat1"),
```
marco-c, line 42: What is the point of this mapping? E.g. can't you print `TranslationUnitRepeat1` instead of `translation_unit_repeat1`?

Luni-4: To be consistent with the `tree-sitter` node notation and furthermore not always we have the same map, for example, https://github.com/mozilla/rust-code-analysis/pull/277/commits/2abb65d1af37cf6f8ee741c3637b89a6f7bc3d20#diff-613041e59dea9037da5917b4337c372965eea13ba7e1e840e0bf765d00f37079R68

Luni-4: Thanks to the help of @calixteman, we have adopted a simple `match` structure to retrieve a string associated to an AST node

marco-c:
Could you run a benchmark (e.g. parse the entire m-c before/after) to see if there is a performance difference?

Luni-4:
> Could you run a benchmark (e.g. parse the entire m-c before/after) to see if there is a performance difference?

Using the following command in debug mode `time ./target/debug/rust-code-analysis-cli -p ../gecko-dev/`:

Before:
```
Executed in    5.16 secs   fish           external
   usr time   50.91 secs  614.00 micros   50.91 secs
   sys time    6.52 secs  287.00 micros    6.52 secs
```
After:

```
Executed in    5.18 secs   fish           external
   usr time   50.67 secs  611.00 micros   50.67 secs
   sys time    6.42 secs  279.00 micros    6.42 secs
```

Luni-4:
> Could you run a benchmark (e.g. parse the entire m-c before/after) to see if there is a performance difference?

I used the following command to test the time difference: `time -o output_file ./target/debug/rust-code-analysis-cli -m -p gecko-dev/ &> /dev/null`

Before:

```
3140.22 
user 39.23
system 5:33.37 
elapsed 953%
CPU (0avgtext+0avgdata 2661084maxresident)k0inputs+0outputs (0major+872686minor)
pagefaults 0swaps
```

After:

```
3160.63
user 40.21
system 5:37.87
elapsed 947%
CPU (0avgtext+0avgdata 2553792maxresident)k0inputs+0outputs (0major+861493minor)
pagefaults 0swaps
```

marco-c:
OK, there is almost no difference then.

marco-c (approved):

marco-c:
Since we are not using phf anymore, could you file an issue about removing the dependency or moving it to the build dependencies?

PR #290, opened 2020-10-09
Test rust-code-analysis-cli on mozilla-central

### LINKED ISSUE #39: Add an integration test parsing all files from mozilla-central

Luni-4:
This PR fixes #39 

Thanks in advance for your review!

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -53,6 +53,37 @@ tasks:
           owner: cdenizet@mozilla.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+      - taskId: {$eval: as_slugid("test_mozilla_central")}
+        created: {$fromNow: ''}
+        deadline: {$fromNow: '1 hour'}
+        provisionerId: proj-relman
+        workerType: ci
+        payload:
+          maxRunTime: 3600
+          image: "rust:buster"
+          command:
+            - "/bin/bash"
+            - "-cx"
+            - "git clone --recursive --quiet ${repository} &&
+               [ ! -d \"cache/gecko-dev\" ] &&
+               git clone --quiet https://github.com/mozilla/gecko-dev.git /cache/gecko-dev || true &&
+               pushd /cache/gecko-dev && git pull origin master && popd &&
+               mkdir -p /tmp/mozilla_central_output &&
+               cd rust-code-analysis &&
+               git -c advice.detachedHead=false checkout ${head_rev} &&
+               cargo build --release --all --all-features &&
+               cargo run --release -p rust-code-analysis-cli -- -p /cache/gecko-dev \
+                                      -j24 --metrics -O json -o /tmp/mozilla_central_output"
+          cache:
+            rust-code-analysis-mozilla-central-repository: /cache
+        scopes:
+          - "docker-worker:cache:rust-code-analysis-mozilla-central-repository"
+        metadata:
+          name: rust-code-analysis mozilla-central test
+          description: rust-code-analysis-cli tested on mozilla-central
+          owner: cdenizet@mozilla.com
+          source: ${repository}/raw/${head_rev}/.taskcluster.yml
+
       - taskId: {$eval: as_slugid("build_documentation")}
         created: {$fromNow: ''}
         deadline: {$fromNow: '2 hour'}
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -53,37 +53,6 @@ tasks:
           owner: cdenizet@mozilla.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
-      - taskId: {$eval: as_slugid("test_mozilla_central")}
-        created: {$fromNow: ''}
-        deadline: {$fromNow: '1 hour'}
-        provisionerId: proj-relman
-        workerType: ci
-        payload:
-          maxRunTime: 3600
-          image: "rust:buster"
-          command:
-            - "/bin/bash"
-            - "-cx"
-            - "git clone --recursive --quiet ${repository} &&
-               [ ! -d \"cache/gecko-dev\" ] &&
-               git clone --quiet https://github.com/mozilla/gecko-dev.git /cache/gecko-dev || true &&
-               pushd /cache/gecko-dev && git pull origin master && popd &&
-               mkdir -p /tmp/mozilla_central_output &&
-               cd rust-code-analysis &&
-               git -c advice.detachedHead=false checkout ${head_rev} &&
-               cargo build --release --all --all-features &&
-               cargo run --release -p rust-code-analysis-cli -- -p /cache/gecko-dev \
-                                      -j24 --metrics -O json -o /tmp/mozilla_central_output"
-          cache:
-            rust-code-analysis-mozilla-central-repository: /cache
-        scopes:
-          - "docker-worker:cache:rust-code-analysis-mozilla-central-repository"
-        metadata:
-          name: rust-code-analysis mozilla-central test
-          description: rust-code-analysis-cli tested on mozilla-central
-          owner: cdenizet@mozilla.com
-          source: ${repository}/raw/${head_rev}/.taskcluster.yml
-
       - taskId: {$eval: as_slugid("build_documentation")}
         created: {$fromNow: ''}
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -53,6 +56,31 @@ tasks:
           owner: cdenizet@mozilla.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+      - taskId: {$eval: as_slugid("test_mozilla_central")}
```
marco-c, line 59: We should use a cache for the repository clone, so we don't reclone from scratch every time.
Here's an example: https://github.com/mozilla/bugbug/blob/ff9f8a78d1f47d2067c130fe5a791a360fac7413/infra/data-pipeline.yml#L13-L47.

Luni-4: I tried to add scopes, but I get the following error: https://github.com/mozilla/rust-code-analysis/commit/aaf9eede020b6212493e3639be156af6da35033f#commitcomment-43107625

marco-c: Need to wait on https://github.com/mozilla/community-tc-config/pull/352.

.taskcluster.yml
```diff
@@ -53,6 +56,33 @@ tasks:
...
+        provisionerId: proj-relman
+        workerType: ci
+        payload:
+          maxRunTime: 3600
+          image: "rust:buster"
+          command:
+            - "/bin/bash"
+            - "-cx"
+            - "git clone --recursive --quiet ${repository} &&
+               git clone --depth 1 https://github.com/mozilla/${mozilla_central}.git &&
```
marco-c, line 71: Nit: also make this quiet

.taskcluster.yml
```diff
@@ -25,6 +25,9 @@ tasks:
       $if: 'tasks_for == "github-pull-request"'
       then: ${event.pull_request.head.repo.html_url}
       else: ${event.repository.html_url}
+
+    mozilla_central: gecko-dev
+    mozilla_central_output: /tmp/mozilla_central_output
```
marco-c, line 30: Let's hardcode the directories in the task itself, it's not worth defining them separately far from the task

.taskcluster.yml
```diff
@@ -53,6 +56,35 @@ tasks:
...
+        provisionerId: proj-relman
+        workerType: ci
+        payload:
+          maxRunTime: 3600
+          image: "rust:buster"
+          command:
+            - "/bin/bash"
+            - "-cx"
+            - "git clone --recursive --quiet ${repository} &&
+               git clone --depth 1 --quiet https://github.com/mozilla/${mozilla_central}.git &&
```
marco-c, line 71: This command should fail when the repo is already in the cache (`fatal: il percorso di destinazione 'rust-code-analysis' esiste già e non è una directory vuota.`).

Luni-4: How can we avoid that?

.taskcluster.yml
Luni-4: How should I define a `let` in a single task?

marco-c: You can just completely avoid defining the variable and just hardcode the actual values (`gecko-dev` and `/tmp/mozilla_central_output`) where they are used

.taskcluster.yml
```diff
@@ -53,6 +54,39 @@ tasks:
...
+          maxRunTime: 3600
+          image: "rust:buster"
+          env:
+            mozilla_central: gecko-dev
+            mozilla_central_output: /tmp/mozilla_central_output
+          command:
+            - "/bin/bash"
+            - "-cx"
+            - "git clone --recursive --quiet ${repository} &&
+               [ ! -d ${mozilla_central} ] &&
```
marco-c, line 72: You can do either: clone if the dir doesn't exist, cd & pull if the dir exists, or clone ignoring errors, then cd & pull.

Luni-4:
@marco-c 

I think there is a cache problem somewhere

marco-c:
> I think there is a cache problem somewhere

Maybe an earlier task failed while cloning the repo, so the directory exists but it is not a valid repository.
You'd need to catch this case too. Writing a little separate script to handle the cloning might be easier than trying to do it all in the task definition.

Luni-4:
> Maybe an earlier task failed while cloning the repo, so the directory exists but it is not a valid repository.
> You'd need to catch this case too. Writing a little separate script to handle the cloning might be easier than trying to do it all in the task definition.

That could be an idea, but I guess it happened because of the wrong commands I set up earlier. In the meanwhile, I tried to delete the empty directory but without success.

marco-c:
Oh wait, I know what's wrong. You are mounting the cache on the "/gecko-dev" path, so "/gecko-dev" will always exist (it's the cache!).
You should have a dir such as "/cache", and then clone the repo inside that directory (so "/cache/gecko-dev").

Luni-4:
Now it seems to work but it doesn't cache anything, it clones the repo every time I force push

Luni-4:
@marco-c 

Is there anything that should be added in this PR? I wasn't able to cache the repository on force push action though

marco-c:
> Is there anything that should be added in this PR? Just a note, I wasn't able to cache the repository on force push, but I guess it's an expected behavior

It should work :S
I'm not sure why it's not!

Anyway, I'm thinking maybe we should only run the task on releases rather than on every push/PR, especially until we manage to properly use the cache. Can you make the change? (also make the deployment task depend on this newly added task).

Luni-4:
> Anyway, I'm thinking maybe we should only run the task on releases rather than on every push/PR, especially until we manage to properly use the cache. Can you make the change? (also make the deployment task depend on this newly added task).

Done

.taskcluster.yml
```diff
@@ -53,6 +53,7 @@ tasks:
           owner: cdenizet@mozilla.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+
```
marco-c, line 56: There is an extraneous newline here, can you remove it?

.taskcluster.yml
```diff
@@ -147,12 +148,49 @@ tasks:
...
+            owner: cdenizet@mozilla.com
+            source: ${repository}/raw/${head_rev}/.taskcluster.yml
+
       - $if: 'tasks_for == "github-push" && head_branch[:10] == "refs/tags/"'
         then:
           taskId: {$eval: as_slugid("build_linux_release")}
           dependencies:
             - {$eval: as_slugid("lint_test_task")}
             - {$eval: as_slugid("windows_test_task")}
+            - {$eval: as_slugid("test_mozilla_central")}
```
marco-c, line 193: We can build the release if that test fails, let's make a dependency of the deployment task.

Luni-4: I thought to add that dependency there to make the CI less heavy. If the `mozilla-central` task fails, we don't create the release binaries at all

marco-c: This test is so slow that the runtime of the others can be considered negligible. Better to catch all errors together, than fixing one, then releasing and noticing a second.

Luni-4: Ok, I'm going to add that check within the deployment task

marco-c (dismissed):

.taskcluster.yml
```diff
@@ -147,6 +147,42 @@ tasks:
...
+              - "git clone --recursive --quiet ${repository} &&
+                 [ ! -d \"cache/gecko-dev\" ] &&
+                 git clone --quiet https://github.com/mozilla/gecko-dev.git /cache/gecko-dev || true &&
+                 pushd /cache/gecko-dev && git pull origin master && popd &&
+                 mkdir -p /tmp/mozilla_central_output &&
+                 cd rust-code-analysis &&
+                 git -c advice.detachedHead=false checkout ${head_rev} &&
+                 cargo build --release --all --all-features &&
+                 cargo run --release -p rust-code-analysis-cli -- -p /cache/gecko-dev \
+                                     -j24 --metrics -O json -o /tmp/mozilla_central_output"
```
marco-c, line 175:
```suggestion
                                     -j4 --metrics -O json -o /tmp/mozilla_central_output"
```

`j24` is probably too much, the `ci` machine is small (`n1-standard-4`), so let's use `j4` instead.

marco-c (approved):

PR #307, opened 2020-10-28
Add a new line feed to the end of an output format

Luni-4:
This PR adds a new line feed to the end of an output format

### DIFF
--- a/rust-code-analysis-cli/src/formats.rs
+++ b/rust-code-analysis-cli/src/formats.rs
@@ -42,7 +42,7 @@ impl Format {
                     } else {
                         serde_json::to_string(&space).unwrap()
                     };
-                    write!(stdout, "{}", json_data)
+                    writeln!(stdout, "{}", json_data)
                 }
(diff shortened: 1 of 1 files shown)


Luni-4:
@marco-c 

Can we merge this one?

marco-c:
Why do you want to add a newline at the end of the output?

Luni-4:
In this way distinct files are printed on different lines and it is much easier to retrieve them

rust-code-analysis-cli/src/formats.rs
```diff
@@ -42,17 +42,17 @@ impl Format {
                     } else {
                         serde_json::to_string(&space).unwrap()
                     };
-                    write!(stdout, "{}", json_data)
+                    write!(stdout, "{}\n", json_data)
```
marco-c, line 45: You can use `writeln!` instead

marco-c (approved):
There is a compilation error, but it is not your fault.

PR #331, opened 2020-11-10
Fix lloc computation for Rust language

### LINKED ISSUE #405: Wrong LLOC computation for the Rust language
Initially we assumed that **ALL** Rust expressions were statements. For example:

Accessing a field --> `struct.field`
Parenthesized expressions --> `(42 + 0)`
Tuple expressions ---> `(0, 42)`
Array expressions ---> `[0; 42]`
Functions and macros when they are not used as statements ---> `if foo() {}` or `if foo!() {}`

This approach leaded to have an high and wrong `LLOC`  for the `Rust` language. It would be helpful to remove these statements and consider only the effective ones.

Luni-4:
This PR fixes #405.

The tests added to this PR avoid repeating the same error in the future.

### DIFF
--- a/src/metrics/loc.rs
+++ b/src/metrics/loc.rs
@@ -310,7 +310,6 @@ impl Loc for RustCode {
             | AssignmentExpression
             | CompoundAssignmentExpr
             | ReturnExpression
-            | TupleExpression
             | UnitExpression
             | IfExpression
             | IfLetExpression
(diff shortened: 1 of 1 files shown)


marco-c (approved):
This looks reasonable to me, but I'd like @calixteman's opinion as he has more experience with this code.

calixteman:
The problem with this PR is that I've no idea on what you're trying to fix.
It isn't only for me but for people (now and/or future) who want to follow what we're doing.
So please file a bug with what's wrong and then mention this PR.

Luni-4:
@calixteman 

Done, opened #405

src/metrics/loc.rs
```diff
@@ -562,6 +573,135 @@ mod tests {
...
+            loc,
+            [(lloc, 2, usize)]
+        );
+    }
+
+    #[test]
+    fn rust_closure_expression_lloc() {
+        check_metrics!(
+            "let a = |i: i32| -> i32 { i + 1 }; // +1
+             a(42);
```
marco-c, line 696: Nit: `// +1` missing here

marco-c (dismissed):

marco-c (approved):

PR #333, opened 2020-11-11
Add nexits average

Luni-4:
Average `NExit` over the total number of functions/closures in a space.


marco-c:
This needs to be rebased

Luni-4:
Done

src/metrics/exit.rs
```diff
@@ -121,3 +140,27 @@ impl Exit for JavaCode {}
...
 impl Exit for CssCode {}
 impl Exit for HtmlCode {}
+
+#[cfg(test)]
+mod tests {
+    use std::path::PathBuf;
+
+    use super::*;
+
+    #[test]
```
marco-c, line 150: Can you also add a "basic" test with a single function? And one with two functions which are not nested?

src/metrics/exit.rs
```diff
@@ -121,3 +140,27 @@ impl Exit for JavaCode {}
...
+            "def f(a, b):
+                 def foo(a):
+                     if a:
+                         return 1
+                 bar = lambda a: lambda b: b or True or True
+                 return bar(foo(a))(a)",
+            "foo.py",
+            PythonParser,
+            nexits,
+            [(exit, 2, usize)],
```
marco-c, line 162: Shouldn't there be 4 exits here in total (one for `f`, one for `foo`, one for each lambda)?

Luni-4: We don't consider closures exit points for now. Perhaps we should add them in another PR

marco-c: Can you file an issue about that, so we don't forget?

marco-c (approved):

src/metrics/exit.rs
marco-c: https://github.com/mozilla/rust-code-analysis/issues/341

PR #343, opened 2020-11-12
Bump tree-sitter-python

Luni-4:
I followed the instructions and run `cargo run -- -lrust -o ../src/languages` in `enums`. I hope it is the correct result


Luni-4:
Closed in favor of #464

PR #344, opened 2020-11-12
Add a script to update tree-sitter submodules

Luni-4:


update-sumbodules.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Update tree-sitter submodules
+#
+# Usage: ./update-sumbodules.sh $tree-sitter-language
+
+# Update submodule
+git submodule update --remote $1
+
+# Recreate the language
+cd enums
```
marco-c, line 11: Nit: use `pushd enums` instead of this, and `popd` instead of `cd ..`, so if the command fails you are not placed in `enums` and have to do `cd ..` manually

Luni-4: We need to set `bash` not `sh` though, because `pushd` and `popd` are bash built-in

update-sumbodules.sh
```diff
@@ -0,0 +1,16 @@
...
+cargo run -- -lrust -o ../src/languages
```
marco-c, line 12: The target language shouldn't always be `rust`

marco-c (dismissed):

update-sumbodules.sh
marco-c: It's a "nit", so feel free to either apply the change or not

marco-c (approved):

PR #345, opened 2020-11-12
Bump tree-sitter-c-sharp

Luni-4:


marco-c (changes requested):
There should be some changes required in src/languages/language_csharp.rs.

PR #346, opened 2020-11-12
Bump tree-sitter-rust

Luni-4:


Luni-4:
Closed in favor of #552

PR #349, opened 2020-11-12
Bump tree-sitter-java

Luni-4:


Luni-4:
Java does not have any metric implementations yet, so we can merge it

marco-c (approved):

PR #363, opened 2020-11-18
Check validity of a tree-sitter-language update

### LINKED ISSUE #361: Write a script to help with updating tree-sitter dependencies
When updating tree-sitter dependencies, there might be changes to the grammars that we might need to take into account.
Sometimes they are new features, so they don't break our current code. Sometimes they might contain breaking changes too, so we need a way to validate them.

We could write a script that:
1) Clones a big repository such as mozilla-central;
2) Runs RCA, storing the metrics in JSON files;
3) Applies the tree-sitter language update;
4) Runs RCA again, storing the metrics in a separate set of JSON files;
5) Generates an explorable diff of the JSON files.

If the diff is empty, or all the changes in the diff makes sense, then we can apply the tree-sitter update.
If some of the changes in the diff are bad, we can't apply the tree-sitter update and we need to debug what's wrong.

Luni-4:
Add a script to check whether a tree-sitter-language update introduces
breaking changes in `rust-code-analysis` code.

To do so, this script comapres the metrics, computed on a chosen
repository, before and after a tree-sitter-language update.

I've used `json-diff` from `npm` for the comparisons because it allows to show the differences using colors, if it exists another `json-diff` in Python or in any other language which can do that I would be happy to change that.

This PR fixes #361

### DIFF
--- a/check-submodule.py
+++ b/check-submodule.py
@@ -1,0 +1,315 @@
+#!/usr/bin/env python3
+
+"""check-submodule
+This script checks whether breaking changes could be introduced in
+rust-code-analysis code after the update of a tree-sitter-language submodule.
+To do so, it compares the differences between the metrics, computed on a
+chosen repository, before and after a tree-sitter-language update.
+
+
+To compute metrics:
+
+./check-submodule.py compute-metrics -u REPO_URL -p LOCAL_DIR -l TREE_SITTER_LANGUAGE
+
+NOTE: The compute-metrics subcommand MUST be run on a clean master branch!
+
+
+To compare metrics:
+
+The metrics are saved as json files, and to compare them, a specific
+json-diff has been adopted.
+
+1. Install npm on your system
+2. Install json-diff from npm running: npm install -g json-diff
+3. Install deepdiff to retrieve minimal tests: pip install deepdiff
+
+./check-submodule.py compare-metrics -l TREE_SITTER_LANGUAGE
+"""
+
+import argparse
+import json
+import pathlib
+import re
+import subprocess
+import sys
+import typing as T
+
+import deepdiff
+
+# The /tmp directory will be used as workdir
+WORKDIR = pathlib.Path("/tmp")
+# Suffix for the directory containing the old metrics
+OLD_SUFFIX = "-old"
+# Suffix for the directory containing the new metrics
+NEW_SUFFIX = "-new"
+
+# Run a subprocess.
+def run_subprocess(cmd: str, *args: T.Union[str, pathlib.Path]) -> None:
+    subprocess.run([cmd, *args])
+
+
+# Run a subprocess and return its output.
+def get_subprocess_output(
+    cmd: str, *args: T.Union[str, pathlib.Path]
+) -> subprocess.CompletedProcess:
+    return subprocess.run(
+        [cmd, *args],
+        text=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+    )
+
+
+# Run rust-code-analysis on the chosen repository to compute metrics.
+def run_rca(repo_dir: pathlib.Path, output_dir: pathlib.Path) -> None:
+    run_subprocess(
+        "cargo",
+        "run",
+        "--package",
+        "rust-code-analysis-cli",
+        "--",
+        "--metrics",
+        "--output-format=json",
+        "--pr",
+        "-p",
+        repo_dir,
+        "-o",
+        output_dir,
+    )
+
+
+# Find the difference between the two json metric files.
+def get_json_diff(
+    first_file: pathlib.Path, second_file: pathlib.Path
+) -> T.Tuple[T.Dict[str, T.Any], T.Dict[str, T.Any]]:
+    with open(first_file, "r") as input_file:
+        t1 = json.load(input_file)
+
+    with open(second_file, "r") as input_file:
+        t2 = json.load(input_file)
+
+    diff = deepdiff.DeepDiff(t1, t2, ignore_order=True)
+
+    return (t1, diff)
+
+
+# Save the filename and the list of code spans associated to the differences
+# in a dictionary.
+def get_metrics_diff_span(
+    first_json: T.Dict[str, T.Any], diff: T.Dict[str, T.Any]
+) -> T.Dict[str, T.List[T.Tuple[int, int]]]:
+    # Search for this pattern in the differences object
+    prog = re.compile(r"\['spaces'\]\[\d+\]")
+
+    output = {"name": first_json["name"], "spaces_spans": []}
+
+    for value in diff["values_changed"]:
+        val = "".join(prog.findall(value))
+        # Subtracting one because files starts from 0
+        start_line = eval(f'first_json{val}["start_line"]') - 1
+        end_line = eval(f'first_json{val}["end_line"]')
+        output["spaces_spans"].append((start_line, end_line))
+
+    # Print the path of the repository file containing the differences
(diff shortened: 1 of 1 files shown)


compare-metrics.py
```diff
@@ -0,0 +1,76 @@
...
+    new_paths = pathlib.Path(new_dir).glob("**/*.json")
+
+    for old_filename, new_filename in zip(old_paths, new_paths):
+        ret_value = run_subprocess("json-diff", "-C", old_filename, new_filename)
+        if ret_value.returncode != 0:
+            print(ret_value.stdout)
+            input_value = input("Continue comparison (y/n)? ")
+            if input_value == "n":
+                sys.exit(0)
+            print()
```
marco-c, line 72: It might be better to show the diffs in an "explorable way".
That is, start with showing the list of directories in which there are differences, let the user navigate within the directories down to the actual files where there are differences, and then down to the functions where there are differences.
This might be easier when there are many differences.

We can do this later though.

compute-metrics.py
```diff
@@ -0,0 +1,113 @@
...
+    # New metrics directory
+    new_dir = WORKDIR / (args.path + "-new")
+
+    # Create output directories
+    old_dir.mkdir(parents=True, exist_ok=True)
+    new_dir.mkdir(parents=True, exist_ok=True)
+
+    # Git clone the chosen repository
+    # Note: no submodules repositories are accepted
+    print("Cloning {} into {}".format(args.url, repo_dir))
```
marco-c, line 92: Nit: here and in other prints, use f-strings:
```Python
print(f"Cloning {args.url} into {repo_dir}")
```

compare-metrics.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python3
```
marco-c, line 1: nit: Since the scripts are so related to each other, can you merge them in a single script and decide the operation (compute or compare) with an argument?

compare-metrics.py
Luni-4: Now there are two modes:

- The `interactive` mode prints on screen the differences among metrics of each file, asking a user if he wants to continue the comparison
- The `compare` mode dumps the json files of differences in a specific directory and prints on screen the paths of files containing the differences.

Luni-4: Removed `interactive` mode because on many files it's pretty cumbersome to use

Luni-4:
- Added a way to retrieve minimal tests from source code
- Refactored and documented the code a bit

check-submodule.py
```diff
@@ -0,0 +1,316 @@
+#!/usr/bin/env python3
+
+"""check-submodule
+This script checks whether breaking changes could be introduced in
+rust-code-analysis code updating a tree-sitter-language submodule.
```
marco-c, line 5:
```suggestion
rust-code-analysis code after the update of a tree-sitter-language submodule.
```

check-submodule.py
```diff
@@ -0,0 +1,316 @@
...
+        "--output-format=json",
+        "--pr",
+        "-p",
+        repo_dir,
+        "-o",
+        output_dir,
+    )
+
+
+# Make the difference between the two json metric files
```
marco-c, line 79:
```suggestion
# Find the difference between the two json metric files
```

check-submodule.py
```diff
@@ -0,0 +1,316 @@
...
+    with open(second_file, "r") as input_file:
+        t2 = json.load(input_file)
+
+    diff = deepdiff.DeepDiff(t1, t2, ignore_order=True)
+
+    return (t1, diff)
+
+
+# Save in a dictionary the filename and the list of code spans associated to
+# the differences
```
marco-c, line 95:
```suggestion
# Save the filename and the list of code spans associated to
# the differences in a dictionary.
```

check-submodule.py
```diff
@@ -0,0 +1,316 @@
...
+# Suffix for the directory containing the new metrics
+NEW_SUFFIX = "-new"
+
+# Run a subprocess
+def run_subprocess(cmd: str, *args: T.Any) -> None:
+    subprocess.run([cmd, *args])
+
+
+# Run a subprocess and return its output
+def get_subprocess_output(cmd: str, *args: T.Any) -> subprocess.CompletedProcess:
```
marco-c, line 52:
```suggestion
def get_subprocess_output(cmd: str, *args: T.Union[str, pathlibPath]) -> subprocess.CompletedProcess:
```

check-submodule.py
```diff
@@ -0,0 +1,316 @@
...
+
+# The /tmp directory will be used as workdir
+WORKDIR = pathlib.Path("/tmp")
+# Suffix for the directory containing the old metrics
+OLD_SUFFIX = "-old"
+# Suffix for the directory containing the new metrics
+NEW_SUFFIX = "-new"
+
+# Run a subprocess
+def run_subprocess(cmd: str, *args: T.Any) -> None:
```
marco-c, line 47:
```suggestion
def run_subprocess(cmd: str, *args: T.Union[str, pathlib.Path]) -> None:
```

check-submodule.py
```diff
@@ -0,0 +1,316 @@
...
+
+# Compute minimal tests
+def compute_minimal_tests(
+    old_filename: pathlib.Path, new_filename: pathlib.Path, compare_dir: pathlib.Path
+) -> None:
+    # Make the difference between the two json file with the aim of
+    # getting the minimal tests
+    first_json, diff = get_json_diff(old_filename, new_filename)
+
+    # Retireve the code spans associated to the differences
```
marco-c, line 162:
```suggestion
    # Retrieve the code spans associated to the differences
```

check-submodule.py
```diff
@@ -0,0 +1,316 @@
...
+    output_path = compare_dir / new_filename.name
+    with open(output_path, "w") as output_file:
+        json.dump(diff_json, output_file, indent=2)
+
+
+# Compute minimal tests
+def compute_minimal_tests(
+    old_filename: pathlib.Path, new_filename: pathlib.Path, compare_dir: pathlib.Path
+) -> None:
+    # Make the difference between the two json file with the aim of
```
marco-c, line 158:
```suggestion
    # Find the difference between the two json file with the aim of
```

check-submodule.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+# Dump minimal tests code in an output file
+def dump_minimal_tests(
+    code_spans_object: T.Dict[str, T.List[T.Tuple[int, int]]],
+    new_filename: pathlib.Path,
+    compare_dir: pathlib.Path,
+) -> None:
+    # Remove duplicates from the list of spans
+    spans_list = list(dict.fromkeys(code_spans_object["spaces_spans"]))
```
marco-c, line 124:
```suggestion
    spans_list = dict.fromkeys(code_spans_object["spaces_spans"])
```

no need to collect into a list

check-submodule.py
```diff
@@ -0,0 +1,316 @@
...
+def dump_json_file(
+    stdout: str, new_filename: pathlib.Path, compare_dir: pathlib.Path
+) -> None:
+    # Load the json file of differences as an object
+    diff_json = json.loads(stdout)
+
+    # Dump json file of differences
+    output_path = compare_dir / new_filename.name
+    with open(output_path, "w") as output_file:
+        json.dump(diff_json, output_file, indent=2)
```
marco-c, line 151: You're parsing stdout and then dumping it again to json, why not writing the file directly instead?

marco-c (approved):

marco-c (approved):

PR #371, opened 2020-11-23
Speed up tree-sitter-language update script

Luni-4:
- Add an option to compute only new metrics
- Compute only the programming language associated to the `tree-sitter-language`

### DIFF
--- a/check-submodule.py
+++ b/check-submodule.py
@@ -203,22 +203,25 @@ def compute_metrics(args: argparse.Namespace) -> None:
     old_dir.mkdir(parents=True, exist_ok=True)
     new_dir.mkdir(parents=True, exist_ok=True)
 
-    # Git clone the chosen repository
-    # Note: no submodules repositories are accepted
-    print(f"Cloning {args.url} into {repo_dir}")
-    run_subprocess("git", "clone", "--depth=1", args.url, repo_dir)
+    # Skip if only new metrics are requested
+    if not args.only_new:
 
-    # Compute old metrics
-    print("\nComputing metrics before the update and saving them in", old_dir)
-    run_rca(repo_dir, old_dir)
+        # Git clone the chosen repository
+        # Note: no submodules repositories are accepted
+        print(f"Cloning {args.url} into {repo_dir}")
+        run_subprocess("git", "clone", "--depth=1", args.url, repo_dir)
 
-    # Create a new branch
-    print("\nCreate a new branch called", args.language)
-    run_subprocess("git", "checkout", "-B", args.language)
+        # Compute old metrics
+        print("\nComputing metrics before the update and saving them in", old_dir)
+        run_rca(repo_dir, old_dir)
 
-    # Update tree-sitter-language submodule
-    print("\nUpdate", args.language)
-    run_subprocess("./update-sumbodules.sh", args.language)
+        # Create a new branch
+        print("\nCreate a new branch called", args.language)
+        run_subprocess("git", "checkout", "-B", args.language)
+
+        # Update tree-sitter-language submodule
+        print("\nUpdate", args.language)
+        run_subprocess("./update-sumbodules.sh", args.language)
 
     # Compute new metrics
     print("\nComputing metrics after the update and saving them in", new_dir)
@@ -259,6 +262,14 @@ def main() -> None:
         "compute-metrics",
         help="Computes the metrics of a chosen repository before and after "
         "a tree-sitter-language update.",
+    )
+
+    # Optional arguments
+    compute_metrics_cmd.add_argument(
+        "--only-new",
+        "-n",
+        action="store_true",
+        help="Only compute the metrics after the tree-sitter-language update",
     )
 
     # Arguments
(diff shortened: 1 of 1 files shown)


check-submodule.py
```diff
@@ -66,23 +68,76 @@
...
+            # Dump json file of differences
+            dump_json_file(json_output, new_filename, self.compare_dir)
+
+            # Compute minimal tests
+            compute_minimal_tests(old_filename, new_filename, self.compare_dir)
+
+
 # Run a subprocess.
 def run_subprocess(cmd: str, *args: T.Union[str, pathlib.Path]) -> None:
     subprocess.run([cmd, *args])
```
marco-c, line 138: I wonder why git shows this as new. Could you move it to the top, so maybe it'll be easier for git to figure out it's still the same?

check-submodule.py
```diff
@@ -66,23 +68,76 @@
...
+        self,
+        old_metrics: T.List[pathlib.Path],
+        new_metrics: T.List[pathlib.Path],
+        compare_dir: pathlib.Path,
+        max_workers: int,
+    ):
+        self.compare_dir = compare_dir
+        self.max_workers = max_workers
+
+        # Max number of filepaths in a sublist
```
marco-c, line 84: Nit: `file paths` or just `paths`

check-submodule.py
```diff
@@ -66,23 +68,76 @@
...
+        n = math.ceil(len(old_metrics) / max_workers)
+
+        # Assign a certain number of filepaths to each worker
+        self.workers_filepaths = [
+            zip(old_metrics[i * n : (i + 1) * n], new_metrics[i * n : (i + 1) * n])
+            for i in range((len(old_metrics) + n - 1) // n)
```
marco-c, line 90: Nit: you can do this more easily if you use the third argument of range, something like `range(0, len(old_metrics), n)`.

>>> list(range(0, 100, 5))
[0, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55, 60, 65, 70, 75, 80, 85, 90, 95]

check-submodule.py
```diff
@@ -66,23 +68,76 @@
...
+        ]
+
+    # Run asynchronous comparisons between json files.
+    async def diff(self):
+        # Define the max number of coroutines used to compare json files
+        all_coroutines = asyncio.gather(
```
marco-c, line 96: Nit: don't define all_coroutines, just do `await asyncio.gather` directly

check-submodule.py
```diff
@@ -66,23 +68,76 @@
...
+            *[
```
marco-c, line 97: Nit: try *(...) instead of *[...] to avoid creating a list

check-submodule.py
```diff
@@ -43,6 +43,29 @@
 # Suffix for the directory containing the new metrics
 NEW_SUFFIX = "-new"
 
+# Extensions parsed by each tree-sitter-language
+EXTENSIONS = {
+    "tree-sitter-mozjs": ["*.js", "*.js2"],
```
marco-c, line 50: `.jsm` too

check-submodule.py
```diff
@@ -43,6 +43,29 @@
...
+    "tree-sitter-javascript": ["*.js", "*.js2"],
+    "tree-sitter-tsx": ["*.tsx"],
+    "tree-sitter-typescript": ["*.ts", "*.jsw", "*.jsmw"],
+    "tree-sitter-java": ["*.java"],
+    "tree-sitter-rust": ["*.rs"],
+    "tree-sitter-cpp": [
```
marco-c, line 56: `cx` too

check-submodule.py
```diff
@@ -203,22 +203,25 @@ def compute_metrics(args: argparse.Namespace) -> None:
     old_dir.mkdir(parents=True, exist_ok=True)
     new_dir.mkdir(parents=True, exist_ok=True)
 
-    # Git clone the chosen repository
-    # Note: no submodules repositories are accepted
-    print(f"Cloning {args.url} into {repo_dir}")
-    run_subprocess("git", "clone", "--depth=1", args.url, repo_dir)
+    # Skip whether only new metrics are requested
```
marco-c, line 287: `Skip when only new metrics are requested` or `Skip if only new metrics are requested`

check-submodule.py
```diff
@@ -261,6 +264,14 @@ def main() -> None:
         "a tree-sitter-language update.",
     )
 
+    # Optional arguments
+    compute_metrics_cmd.add_argument(
+        "--new",
+        "-n",
+        action="store_true",
+        help="Compute only the metrics after the tree-sitter-language update",
```
marco-c, line 353: `Only compute the metrics after the tree-sitter-language update`

check-submodule.py
```diff
@@ -261,6 +264,14 @@ def main() -> None:
         "a tree-sitter-language update.",
     )
 
+    # Optional arguments
+    compute_metrics_cmd.add_argument(
+        "--new",
```
marco-c, line 350: Let's call it `only-new` so it's more clear

Luni-4:
@marco-c 

Let's address the asynchronous part in another PR, let's merge only these two commits for now

marco-c (approved):

PR #373, opened 2020-11-26
Fix JavaScript nom

Luni-4:
We were not counting closures in the correct way

### DIFF
--- a/src/metrics/nom.rs
+++ b/src/metrics/nom.rs
@@ -68,6 +68,105 @@ impl Stats {
     }
 }
 
+#[inline(always)]
+fn is_child(node: &Node, id: u16) -> bool {
+    node.object()
+        .children(&mut node.object().walk())
+        .any(|child| child.kind_id() == id)
+}
+
+#[inline(always)]
+fn has_sibling(node: &Node, id: u16) -> bool {
+    let parent = node.object().parent();
+    if let Some(parent) = parent {
+        node.object()
+            .children(&mut parent.walk())
+            .any(|child| child.kind_id() == id)
+    } else {
+        false
+    }
+}
+
+macro_rules! function {
+    ($node: ident, $stats: ident) => {
+        // If a function node has named ancestors or child then it is a function,
+        // otherwise a closure.
+        if count_specific_ancestors!(
+            $node,
+            VariableDeclarator | AssignmentExpression | LabeledStatement | Pair,
+            StatementBlock | ReturnStatement | NewExpression | Arguments
+        ) > 0
+            || is_child($node, Identifier as u16)
+        {
+            $stats.functions += 1;
+        } else {
+            $stats.closures += 1;
+        }
+    };
+}
+
+macro_rules! arrow_function {
+    ($node: ident, $stats: ident) => {
+        // If an arrow function node has named ancestors then it is a function,
+        // otherwise a closure.
+        if count_specific_ancestors!(
+            $node,
+            VariableDeclarator | AssignmentExpression | LabeledStatement,
+            StatementBlock | ReturnStatement | NewExpression | CallExpression
+        ) > 0
+            || has_sibling($node, PropertyIdentifier as u16)
+        {
+            $stats.functions += 1;
+        } else {
+            $stats.closures += 1;
+        }
+    };
+}
+
+macro_rules! js_grammar {
+    ($grammar: ident, $node: ident, $stats: ident) => {
+        use $grammar::*;
+
+        match $node.object().kind_id().into() {
+            FunctionDeclaration | MethodDefinition => {
+                $stats.functions += 1;
+            }
+            GeneratorFunction | GeneratorFunctionDeclaration => {
+                $stats.closures += 1;
+            }
+            Function => {
+                function!($node, $stats);
+            }
+            ArrowFunction => {
+                arrow_function!($node, $stats);
+            }
+            _ => {}
+        }
+    };
+}
+
+macro_rules! typescript_grammar {
+    ($grammar: ident, $node: ident, $stats: ident) => {
+        use $grammar::*;
+
+        match $node.object().kind_id().into() {
+            FunctionDeclaration | MethodDefinition => {
+                $stats.functions += 1;
+            }
+            GeneratorFunction | GeneratorFunctionDeclaration => {
+                $stats.closures += 1;
+            }
+            Function => {
+                function!($node, $stats);
+            }
+            ArrowFunction => {
+                arrow_function!($node, $stats);
+            }
+            _ => {}
+        }
+    };
+}
+
 #[doc(hidden)]
 pub trait Nom
 where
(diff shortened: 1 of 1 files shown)


src/metrics/nom.rs
```diff
@@ -240,6 +240,25 @@ mod tests {
         );
     }
 
+    #[test]
+    fn c_nom() {
```
marco-c, line 244: git is not smart enough to notice it was just moved, please do the move in a separate commit or PR, otherwise this change will mess up the blame info.

src/metrics/nom.rs
```diff
@@ -260,20 +279,23 @@ mod tests {
...
-            "int foo();
-
-             int foo() {
-                 return 0;
+            "function f(a, b) {
+                 function foo(a) {
+                     return a;
+                 }
+                 var bar = function (a, b) {return a + b};
+                 var bar1 = function (a) {return a};
```
marco-c, line 289: Why considering these differently than `foo`? They are all named functions.
We need to decide on our definition of what a closure is

Luni-4: Because in the `foo` case we have a `FunctionDefinition`, while in the `bar` case we have a function _assigned_ to a variable. I've used this criteria to discriminate among functions and closures

Luni-4: Perhaps we can use this kind of closures for `Javascript`:

```
var add = (function () {
  var counter = 0;
  return function () {counter += 1; return counter}
})();

add();
add();
add();
```

Taken from [here](https://www.w3schools.com/js/js_function_closures.asp)

Luni-4: Reading more carefully the example, in JavaScript, a closure is a mechanism to make global variables as local variables of a function. Do we want to detect this mechanism as a closure?

marco-c: Basically, any JS function forms a closure.
From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures:
> A closure is the combination of a function bundled together (enclosed) with references to its surrounding state (the lexical environment). In other words, a closure gives you access to an outer function’s scope from an inner function. In JavaScript, closures are created every time a function is created, at function creation time.

Maybe we should totally change what we're doing, remove the "closure" concept and define a "inner function" concept instead?
@calixteman what do you think?

Luni-4: Since the definition above describes what a closure is, I propose to maintain the word "closure" and change the implementation instead. Introducing a new word could be confusing. What do you think @marco-c and @calixteman?

marco-c: @Luni-4 in JavaScript any function matches that definition, so if we keep using the word "closure" we'd need to consider any function as a closure for the JS language.

Luni-4: @marco-c So why have functions and closures been splitted in? So a function doesn't exist at all because this property: _In other words, a closure gives you access to an outer function’s scope from an inner function._  is always true?

In that case, let's count only functions, specifying that functions and closures are the same

marco-c: We initially considered a "closure" an "inner function", because in all languages except JS that works well enough. In JS instead any function is a closure, so that assumption is broken.
I wouldn't keep the same word ("closure") now, as I think there is often a logic difference between "top-level functions" and "inner functions".

We could do one of or a mix of:
- separate functions between "named functions" and "unnamed functions" (though often small inner functions have a name);
- separate functions between "top-level functions" and "inner functions" (though this doesn't work with a common pattern of JS like:
```
(function() {
    function ciao() {
        // Ciao
    }
})();
```

```
function ciao() {
    // Ciao
}
```
these two would be considered differently, but I would instead consider them the same).

I don't see a clear winner

calixteman: > Basically, any JS function forms a closure.
> From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures:
> 
> > A closure is the combination of a function bundled together (enclosed) with references to its surrounding state (the lexical environment). In other words, a closure gives you access to an outer function’s scope from an inner function. In JavaScript, closures are created every time a function is created, at function creation time.
> 
> Maybe we should totally change what we're doing, remove the "closure" concept and define a "inner function" concept instead?
> @calixteman what do you think?

I think we don't care so much about detail of implementation.
From my pov, there are 2 kind of functions: named ones and anonymous ones.
Named ones may have different patterns: `function name(...`, ` name = (...) => ...`, `name = (funtion(){ return function()...)())`.
Everything is a kind of "inner function"... so I'm not sure that it's relevant here.

marco-c: This would be the first option in my last comment:
> separate functions between "named functions" and "unnamed functions" (though often small inner functions have a name);

The only drawback is that some developers assign names to all functions, and then we'd lose the differentiation.

Luni-4: To wrap it up, in *JS*:

- Named functions should be treated as functions
- Unnamed functions like `return function()...)()` as closures

Is this the kind of approach we want to follow to?

Luni-4:
More tests need to be added https://community-tc.services.mozilla.com/api/queue/v1/task/X4bywn8bQKmAooULVBsNIA/runs/0/artifacts/public%2Fjson-diffs-and-minimal-tests.tar.gz

Some regressions

Luni-4:
@marco-c 

Fixed every regression, please have a look at the `tar.gz` in the previous message to verify if I haven't done any error

src/metrics/nom.rs
```diff
@@ -97,12 +108,41 @@ impl Nom for MozjsCode {
...
+                    StatementBlock | ReturnStatement | NewExpression | Arguments
+                ) > 0
+                    || is_child_identifier(node, Identifier as u16)
+                {
+                    stats.functions += 1;
+                } else {
+                    stats.closures += 1;
+                }
+            }
+            ArrowFunction => {
```
marco-c, line 132: What if you handled ArrowFunction in exactly the same way as Function?

src/metrics/nom.rs
```diff
@@ -113,12 +153,41 @@ impl Nom for JavascriptCode {
...
         match node.object().kind_id().into() {
-            Function | FunctionDeclaration | MethodDefinition => {
+            FunctionDeclaration | MethodDefinition => {
                 stats.functions += 1;
             }
-            GeneratorFunction | GeneratorFunctionDeclaration | ArrowFunction => {
+            GeneratorFunction | GeneratorFunctionDeclaration => {
                 stats.closures += 1;
             }
+            Function => {
```
marco-c, line 162: It's a pain that we have to duplicate this code for mozjs and for js!
Are there actually differences or can we share the same implementation?

src/metrics/nom.rs
```diff
@@ -145,12 +243,41 @@ impl Nom for TsxCode {
         use Tsx::*;
 
         match node.object().kind_id().into() {
-            Function | FunctionDeclaration | MethodDefinition => {
+            FunctionDeclaration | MethodDefinition => {
```
marco-c, line 246: Same for ts and tsx, we should be able to share the impl

src/metrics/nom.rs
```diff
@@ -277,4 +499,87 @@ mod tests {
...
+                (closures, 1, usize),  // materials.map
+                (total, 2, usize)
+            ]
+        );
+    }
+
+    #[test]
+    fn javascript_arrow_assignment_nom() {
+        check_metrics!(
+            "sink.onPull = () => { };",
```
marco-c, line 542: This one is kind of a corner case. I guess we'll never have a fully perfect way to differentiate between functions and unnamed functions

Luni-4: So do you propose to delete that test?

marco-c: No, let's keep it, it was just a thought

src/metrics/nom.rs
Luni-4: `javascript_arrow_new_nom` and `javascript_arrow_nom` fail

marco-c: OK, not feasible then.
Could you move the `count_specific_ancestors!(` part in a helper function, so we don't have to duplicate it?

src/metrics/nom.rs
calixteman: @marco-c every arrow function in the rhs of an assignment is named function or rhs of property definition `{foo: () => ...}`, or do you see any cases where it isn't the case ?

marco-c: @calixteman it is always the case, what I was thinking are cases like:
```
window.onload = function() { blabla }
```
which we consider a named function, because it theoretically is, but in practice it is an unnamed one if you look at it from the perspective of what the developer meant.

src/metrics/nom.rs
```diff
@@ -68,6 +68,96 @@ impl Stats {
     }
 }
 
+#[inline(always)]
+fn is_child_identifier(node: &Node, id: u16) -> bool {
+    let mut cursor = node.object().walk();
+    for child in node.object().children(&mut cursor) {
```
calixteman, line 74: nit: if possible, use `Iterator::any` here.

src/metrics/nom.rs
```diff
@@ -68,6 +68,96 @@ impl Stats {
     }
 }
 
+#[inline(always)]
+fn is_child_identifier(node: &Node, id: u16) -> bool {
```
calixteman, line 72: Why having `identifier` in the function name: it seems unrelated to what the function is doing.

src/metrics/nom.rs
```diff
@@ -68,6 +68,96 @@ impl Stats {
...
+            $stats.functions += 1;
+        } else {
+            $stats.closures += 1;
+        }
+    };
+}
+
+macro_rules! arrow_function {
+    ($node: ident, $stats: ident) => {
+        // Consider named functions as functions and unnamed ones as
```
calixteman, line 102: Same comment as macro above, but the macros are slightly different so the comment should reflected that.

calixteman (dismissed):
r+, with comments addressed and green CI.

marco-c (approved):

PR #374, opened 2020-11-26
Move the web server in a new crate

### LINKED ISSUE #372: rust-code-analysis-cli should not contain the web features
Currently, the rca-cli also ships the web service (based on actix).

As we are talking about a CLI, it should only provide the CLI features.
Instead, we should split the two features and have a rust-code-analysis-web service as a new crate.

Luni-4:
This PR fixes #372 


rust-code-analysis-web/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+[[bin]]
+name = "rust-code-analysis-web"
+
+[dependencies]
+actix-rt = "^1.0"
+actix-web = "^3.2"
+clap = "^2.33"
+futures = "^0.3"
+num_cpus = "^1.13"
+rust-code-analysis = { path = "..", version = "0.0"}
```
marco-c, line 19: Do we need to specify the version here?

rust-code-analysis-cli/src/main.rs
```diff
@@ -25,8 +22,24 @@ use std::{process, thread};
...
+use rust_code_analysis::LANG;
+
+// Structs
+use rust_code_analysis::{
+    CommentRm, CommentRmCfg, Count, CountCfg, Dump, DumpCfg, Find, FindCfg, Function, FunctionCfg,
+    Metrics, MetricsCfg, PreprocParser, PreprocResults,
+};
+
+// Traits
+use rust_code_analysis::ParserTrait;
```
marco-c, line 42: Why splitting them? Let's move them all together under one `use`

marco-c: Actually, let's keep the old code here, the change isn't related to this PR

rust-code-analysis-cli/src/main.rs
```diff
@@ -587,5 +565,3 @@ fn main() {
         }
     }
 }
-
-// cargo run --release -- -p ../mozilla-central.hg/ -j24 -X "**/third_party/rust/**/*.*" -X "**/*.mako.rs" -X "**/js/src/jit/**/*MacroAssembler*" -X "**/js/src/vm/Opcodes.h" -X "**/obj-x86_64-pc-linux-gnu/**/*.*" -i -c
```
marco-c, line 591: If you want to remove this comment, please submit a separate PR as it's not related to this change

rust-code-analysis-cli/src/main.rs
Luni-4: I have problems with `*` sometimes, but you are right, I can open another PR

rust-code-analysis-cli/src/main.rs
```diff
@@ -25,8 +22,8 @@ use std::{process, thread};
 use walkdir::{DirEntry, WalkDir};
 
 use formats::Format;
+
```
marco-c, line 25: Nit: remove extraneous new line

rust-code-analysis-web/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "rust-code-analysis-web"
+version = "0.0.18"
+authors = ["Calixte Denizet <cdenizet@mozilla.com>"]
+edition = "2018"
+keywords = ["metrics"]
```
marco-c, line 6: We should improve these keywords. I've filed a follow-up: #375.

marco-c (approved):

rust-code-analysis-cli/src/main.rs
Luni-4: In a new PR, thanks for catching it! :)

rust-code-analysis-web/Cargo.toml
Luni-4: Thank you! :)

PR #385, opened 2020-11-30
Further speed up test script 

Luni-4:
- Compare metrics asynchronously 
- Remove `json-diff` because it takes a long time to be run on the mozilla-central repository

### DIFF
--- a/check-submodule.py
+++ b/check-submodule.py
@@ -180,14 +180,9 @@ def dump_minimal_tests(
     filename = code_spans_object["name"]
 
     # Read code spans from the input source code
-    with open(filename, "r", encoding="utf-8") as input_file:
-        try:
-            # Decode only utf-8 source code files
-            lines = input_file.readlines()
-        except UnicodeDecodeError:
-            # Print the file containing non-utf-8 characters and return
-            print("Contains non-UTF-8 encoding:", filename)
-            return
+    with open(filename, "r", encoding="utf-8", errors="ignore") as input_file:
+        # Decode only utf-8 source code files
+        lines = input_file.readlines()
 
     # Write spans to output file
     output_path = compare_dir / new_filename.stem
--- a/check-submodule.py
+++ b/check-submodule.py
@@ -27,7 +27,9 @@ json-diff has been adopted.
 """
 
 import argparse
+import asyncio
 import json
+import math
 import pathlib
 import re
 import subprocess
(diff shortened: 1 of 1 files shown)


check-submodule.py
```diff
@@ -67,23 +64,51 @@
...
+        # Define the max number of coroutines used to compare json files
+        await asyncio.gather(
+            *(
+                self._worker(worker_filepaths)
+                for worker_filepaths in self.workers_filepaths
+            )
+        )
+
+    # Save json files of differences and minimal tests in the chosen directory.
+    async def _worker(self, worker_list: T.List[pathlib.Path]):
```
marco-c, line 100: Nit: could be a function defined directly in `diff`

check-submodule.py
```diff
@@ -158,8 +183,14 @@ def dump_minimal_tests(
     filename = code_spans_object["name"]
 
     # Read code spans from the input source code
-    with open(filename, "r") as input_file:
-        lines = input_file.readlines()
+    with open(filename, "r", encoding="utf-8") as input_file:
```
marco-c, line 186: Nit: You could ignore encoding errors and run on those files anyway

marco-c (dismissed):

check-submodule.py
Luni-4: encoding="utf-8" is the default encoding mode used by `Python`, but to avoid problems with the different versions, I stated it explicitly. Or are you advising to convert non-utf-8 files into utf-8?

marco-c: Yep, I know, I was suggesting ignoring encoding errors.
I'm sure you can do that by reading bytes and then `decode('utf-8', 'ignore')`, maybe there is a way to do the same with `open`.

check-submodule.py
```diff
@@ -158,8 +180,14 @@ def dump_minimal_tests(
     filename = code_spans_object["name"]
 
     # Read code spans from the input source code
-    with open(filename, "r") as input_file:
-        lines = input_file.readlines()
+    with open(filename, "r", encoding="utf-8", errors="ignore") as input_file:
+        try:
+            # Decode only utf-8 source code files
+            lines = input_file.readlines()
+        except UnicodeDecodeError:
```
marco-c, line 187: This should no longer happen now that you're ignoring errors, right?

marco-c (dismissed):

check-submodule.py
Luni-4: Forgot to remove those lines :P 

Thanks! :)

marco-c (approved):

PR #386, opened 2020-12-01
Do not produce different files for each run on the mozilla-central repository

Luni-4:
When `rust-code-analysis-cli` was being run on the `mozilla-central` repository many times, the filenames of the output formats produced at each run were different. That was because the old code assigned filenames without considering that the thread scheduler picks up input files every time in a different order.

So, to avoid conflicts among files, we now use input file paths as output format filenames.

### DIFF
--- a/rust-code-analysis-cli/src/formats.rs
+++ b/rust-code-analysis-cli/src/formats.rs
@@ -1,4 +1,3 @@
-use regex::Regex;
 use std::fs::File;
 use std::io::Write;
 use std::io::{Error, ErrorKind};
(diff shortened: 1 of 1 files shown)


rust-code-analysis-cli/src/formats.rs
```diff
@@ -62,22 +62,20 @@ impl Format {
...
 
-            let output_path = output_path.as_ref().unwrap();
+            // Use file path as filename
+            let filename = path.to_str().unwrap();
 
-            let mut file = path.as_path().file_name().unwrap().to_os_string();
-            file.push(format_ext);
+            // Remove ./ ../ .\ ..\ present in a path
+            let re = Regex::new(r"\.\./|\.\.\\|^\./|^\.\\").unwrap();
+            let no_dots_filename = re.replace_all(&filename, "");
```
marco-c, line 70: Is there some std lib function to do something like this?

rust-code-analysis-cli/src/formats.rs
```diff
@@ -62,22 +62,20 @@ impl Format {
...
 
-            if format_path.as_path().exists() {
-                let mut new_filename = path.to_str().unwrap().to_string();
-                let re = Regex::new(r"[\\:/]").unwrap();
-                new_filename = re.replace_all(&new_filename, "_").to_string();
-                new_filename.push_str(format_ext);
-                format_path.pop();
-                format_path.push(new_filename);
-            }
+            // Create a clean filename
```
marco-c, line 75:
```suggestion
            // Create a clean filename by replacing `:` and slashes with `_`
```

rust-code-analysis-cli/src/formats.rs
Luni-4: I think we could obtain the same result, reducing the code at the same time, using the [components](https://doc.rust-lang.org/std/path/struct.Path.html#method.components) function. I have to try though

marco-c (approved):
Looks much nicer now

PR #387, opened 2020-12-03
Fix cognitive metric when there are no functions

Luni-4:


Luni-4:
@marco-c 

I propose to use the 0 approach instead of the `null` one when there are no functions. Since we are using a _counting_ approach, if something doesn't exist, the count remains set to 0

marco-c:
I'm thinking we should use `null`, so users can differentiate an actual 0 cognitive complexity from a not-existent cognitive complexity.
Users who want to consider `null` as 0 can do so, and users who want to differentiate can do that too.
Instead, if we save 0 for everything, only the first kind of usage would be possible.

Luni-4:
Seems good to me, I'm going to change the metrics in that way then

Luni-4:
@marco-c 

Now it is ready

marco-c (approved):

PR #400, opened 2020-12-16
Clean up filenames produced by rust-code-analysis-cli

Luni-4:
In #386, when a path is cleaned up to create a new filename, the root and the current directory symbol haven't been removed, moreover some symbols lead to duplicates filenames. This PR fixes those problems.

### DIFF
--- a/rust-code-analysis-cli/src/formats.rs
+++ b/rust-code-analysis-cli/src/formats.rs
@@ -61,17 +61,23 @@ impl Format {
                 Format::Yaml => ".yml",
             };
 
-            // Remove . / \ .. symbols from a path to create a unique filename
+            // Remove root /
+            let path = path.strip_prefix("/").unwrap_or(path);
+
+            // Remove root ./
+            let path = path.strip_prefix("./").unwrap_or(path);
+
+            // Replace .. symbol with "_" to create a unique filename
             let cleaned_path: Vec<&str> = path
                 .iter()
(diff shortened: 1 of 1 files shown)


rust-code-analysis-cli/src/formats.rs
```diff
@@ -66,7 +66,7 @@ impl Format {
                 .iter()
                 .filter(|v| {
                     if let Some(s) = v.to_str() {
-                        ![".", ".."].contains(&s)
+                        ![".", "..", "/", "\\", ":"].contains(&s)
```
marco-c, line 69: These are path components, how can there be `/` or `\\` in them?

Luni-4: The [components](https://doc.rust-lang.org/std/path/enum.Component.html#examples) function maintains the root of a path, so if we pass an absolute path `\` on Unix and `C:\` on Windows are maintained.

marco-c: We could remove the root first and then get the components of the resulting relative path

Luni-4: Yep :)

Luni-4: It is not necessary to remove Windows symbols to create a filename, so we can only remove the root from Linux paths.

marco-c (approved):

PR #404, opened 2020-12-22
Final submodule script version

### LINKED ISSUE #367: Speed up tree-sitter-language update script
- [ ] Consider only the files associated to a `tree-sitter-language` when `rust-code-analyses` computes the metrics
- [ ] Compare metrics in parallel using the `multiprocess` package

Luni-4:
Depends on #400 

Fixes #367

This PR provide the final script to compute and compare metrics in order to verify whether there are differences among each tree-sitter-language update.

The changes present in this PR are:
- Split the shell script used to update a tree-sitter-language in order to be more suitable for CI systems
- Add a command to compute metrics on CI systems that allows to choose where the repository containing source codes is located
- Use external Rust commands to compare metrics that provide a considerable speed up on large repositories


update-submodule-ci.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
marco-c, line 1: Can you rename this script to something like "update-language-bindings.sh"?

check-submodule.py
```diff
@@ -229,6 +210,82 @@ def save_diff_files(
     asyncio.run(json_diff.diff())
 
 
+# Run rust-code-analysis on the chosen repository to compute metrics.
+def run_rca(
```
marco-c, line 75: Please move this function where it was before, so the diff is smaller and simpler to review

check-submodule.py
```diff
@@ -247,7 +305,7 @@ def compute_metrics(args: argparse.Namespace) -> None:
     old_dir.mkdir(parents=True, exist_ok=True)
     new_dir.mkdir(parents=True, exist_ok=True)
 
-    # Skip if only new metrics are requested
+    # Skip whether only new metrics are requested
```
marco-c, line 169: The previous sentence was correct (even though it doesn't sound great), this one is not.
Maybe use `when` instead of `if`.

Note: this change should be in a separate commit/PR, as it is unrelated to this commit

check-submodule.py
```diff
@@ -238,6 +295,7 @@ def compute_metrics(args: argparse.Namespace) -> None:
 
     # Repository local directory
     repo_dir = WORKDIR / args.path
+
```
marco-c, line 159: Don't add a newline in an unrelated place

check-submodule.py
Luni-4: That was a reminiscence of an old change, sorry :P

check-submodule.py
```diff
@@ -229,6 +238,54 @@ def save_diff_files(
...
+    print(f"Cloning rust-code-analysis master branch into /tmp")
+    run_subprocess(
+        "git",
+        "clone",
+        "--depth=1",
+        "--recurse-submodules",
+        "-j8",
+        "https://github.com/mozilla/rust-code-analysis",
+        rca_path,
+    )
```
marco-c, line 135: We already have a local clone, we can just reuse it by checking out to master instead of recloning.

Luni-4: Some CI systems fetch a single branch and you need to reset commits and make complicated stuff to retrieve master branch's commits. This solution could potentially be used on any CI system.

marco-c (approved):

PR #411, opened 2021-01-08
Verify metrics correctness after a submodule update

### LINKED ISSUE #406: Add a TaskCluster task to automatically detect metrics differences when a tree-sitter submodule is being updated

Luni-4:
Add an automatic way to verify whether metrics are correct after a
tree-sitter language update.

This PR fixes #406

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -147,6 +147,40 @@ tasks:
           owner: cdenizet@mozilla.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+      - taskId: {$eval: as_slugid("check_submodules")}
+        dependencies:
+           - {$eval: as_slugid("lint_test_task")}
+           - {$eval: as_slugid("windows_test_task")}
+        created: {$fromNow: ''}
+        deadline: {$fromNow: '1 hour'}
+        provisionerId: proj-relman
+        workerType: ci
+        payload:
+          maxRunTime: 3600
+          image: "rust:buster"
+          command:
+            - "/bin/bash"
+            - "-cx"
+            - "git clone --recursive --quiet ${repository} &&
+               cd rust-code-analysis &&
+               git -c advice.detachedHead=false checkout --recurse-submodules ${head_rev} &&
+               rustup component add rustfmt &&
+               ./check-submodules.sh"
+          cache:
+            rust-code-analysis-mozilla-central-repository: /cache
+          artifacts:
+            public/json-diffs-and-minimal-tests.tar.gz:
+              expires: {$fromNow: '2 weeks'}
+              path: /tmp/json-diffs-and-minimal-tests.tar.gz
+              type: file
+        scopes:
+          - "docker-worker:cache:rust-code-analysis-mozilla-central-repository"
+        metadata:
+          name: rust-code-analysis check submodules
+          description: rust-code-analysis check submodules
+          owner: cdenizet@mozilla.com
+          source: ${repository}/raw/${head_rev}/.taskcluster.yml
+
       - $if: 'tasks_for == "github-push" && head_branch[:10] == "refs/tags/"'
         then:
           taskId: {$eval: as_slugid("test_mozilla_central")}
--- a/check-submodules.sh
+++ b/check-submodules.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+# Checks out if a submodule has been updated
+SUBMODULES=`git submodule--helper list | awk '{ print $4 }'`
+RUN_CI="no"
+for SUBMODULE in $SUBMODULES
+do
+    git diff --exit-code HEAD^ $SUBMODULE
+    CHANGED=$? # Get git diff exit code
+    if [ $CHANGED -eq 1 ]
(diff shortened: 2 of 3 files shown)


check-submodules.sh
```diff
@@ -0,0 +1,53 @@
...
+    fi
+done
+
+# If no submodule has been updated, exit the script
+if [ "$RUN_CI" = "no" ]; then
+    exit 0
+fi
+
+# Install json-diff
+JSON_DIFF_LINK="https://github.com/Luni-4/json-diff/releases/download"
```
marco-c, line 24: Nit: why don't you publish it on crates.io, so we can add it to our dev-dependencies?

check-submodules.sh
```diff
@@ -0,0 +1,53 @@
...
+JSON_DIFF_VERSION="0.0.1"
+curl -L "$JSON_DIFF_LINK/v$JSON_DIFF_VERSION/json-diff-linux.tar.gz" |
+tar xz -C $CARGO_HOME/bin
+
+# Install json minimal tests
+JMT_LINK="https://github.com/Luni-4/json-minimal-tests/releases/download"
```
marco-c, line 30: Same here

check-submodules.sh
```diff
@@ -0,0 +1,53 @@
...
+JMT_VERSION="0.0.1"
+curl -L "$JMT_LINK/v$JMT_VERSION/json-minimal-tests-linux.tar.gz" |
+tar xz -C $CARGO_HOME/bin
+
+# Download mozilla-central repository
+MOZILLA_CENTRAL_REPO="https://github.com/mozilla/gecko-dev"
+git clone --depth 1 -j8 $MOZILLA_CENTRAL_REPO $HOME/mozilla-central
```
marco-c, line 37: You can reuse the commands we use in the `test_mozilla_central` task to avoid recloning if the repo is already there

.taskcluster.yml
```diff
@@ -147,6 +147,36 @@ tasks:
...
+          artifacts:
+            public/json-diffs-and-minimal-tests.tar.gz:
+              expires: {$fromNow: '2 weeks'}
+              path: /tmp/json-diffs-and-minimal-tests.tar.gz
+              type: file
+        metadata:
+          name: rust-code-analysis check submodules
+          description: rust-code-analysis check submodules
+          owner: cdenizet@mozilla.com
+          source: ${repository}/raw/${head_rev}/.taskcluster.yml
```
marco-c, line 178: Let's add a cache as we do for the `test_mozilla_central` task

check-submodules.sh
Luni-4: I will publish it as soon as I can. Do you want this script to be transformed in a test?

check-submodules.sh
Luni-4: This one won't be published. It's useless for any other user except us. Let's treat it with the { path: ...} notation if we want to add it to a future test

.taskcluster.yml
Luni-4: Cache didn't work the last time

marco-c: Do you mean it doesn't work in `test_mozilla_central`?

check-submodules.sh
marco-c: No it's fine this way

marco-c: A test would be nicer, but it isn't worth spending time on it

check-submodules.sh
marco-c: It's OK, publishing doesn't hurt, indeed there are a lot of packages that are used by very few people. You never know, maybe it'll be useful for other projects of ours.
Anyway, it's a `nit`, so feel free to ignore ;)

.taskcluster.yml
Luni-4: Exactly, there were problems with Taskcluster

marco-c: OK, let's add it anyway as it's just two lines here, then I'll figure out why it's not working (let's file an issue about it so I don't forget)

check-submodules.sh
Luni-4: Updated on crates.io

check-submodules.sh
Luni-4: Updated on crates.io

marco-c (approved):
> Updated on crates.io

Can you use that instead of downloading and extracting the binary from GitHub releases?

Luni-4:
> > Updated on crates.io
> 
> Can you use that instead of downloading and extracting the binary from GitHub releases?

When you run `cargo install` the crate isn't compiled on the machine? That would make everything much slower than download a single binary and put it in a directory.

marco-c:
> When you run cargo install the crate isn't compiled on the machine? That would make everything much slower than download a single binary and put it in a directory.

True, but it makes it much easier to handle dependencies (and at some point we will want to make this a proper Rust test).
Anyway, we can also do it later on, I don't mind.

Luni-4:
Yep, I agree with you. Let's do that later on

marco-c (approved):

PR #413, opened 2021-01-08
Build Linux release binaries with the musl library

Luni-4:
This PR fixes #413.

Using musl avoids that binaries don't run because of some not updated dependencies.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -229,29 +229,29 @@ tasks:
           workerType: ci
           payload:
             maxRunTime: 3600
-            image: "rust:buster"
+            image: "ekidd/rust-musl-builder:stable"
             command:
               - "/bin/bash"
               - "-cx"
               - "git clone --recursive --quiet ${repository} &&
                  cd rust-code-analysis &&
                  git -c advice.detachedHead=false checkout --recurse-submodules ${head_rev} &&
-                 cargo build --workspace --release &&
-                 cargo package --all-features &&
-                 pushd rust-code-analysis-cli && cargo package --all-features && popd &&
-                 pushd rust-code-analysis-web && cargo package --all-features && popd &&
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -233,14 +233,17 @@ tasks:
...
+                 rustup target add x86_64-unknown-linux-musl &&
+                 git clone --recursive --quiet ${repository} &&
                  cd rust-code-analysis &&
                  git -c advice.detachedHead=false checkout --recurse-submodules ${head_rev} &&
-                 cargo build --all --release &&
-                 cargo package --all-features &&
-                 pushd rust-code-analysis-cli && cargo package --all-features && popd &&
-                 pushd rust-code-analysis-web && cargo package --all-features && popd &&
-                 cd target/release &&
+                 env CC=\"gcc\" CXX=\"g++\" cargo build --workspace --release --target x86_64-unknown-linux-musl &&
```
marco-c, line 242: Why do we need to define CC and CXX?

Luni-4: To compile `tree-sitter` submodules, otherwise the `musl` target fails

marco-c: Why does it fail? I mean, why doesn't it automatically use the right compiler? What compiler does it try to use? Do you have an example of a failure?

Luni-4:
To build correctly with the `rust:buster` image, we need to set up `CC` and `CXX` environment variables to find the correct compilers on the system:

`env CC=\"gcc\" CXX=\"g++\" cargo build --workspace --release --target **x86_64-unknown-linux-musl`

As stated by this discussion also: https://github.com/tree-sitter/tree-sitter/discussions/794#discussioncomment-188114

marco-c (approved):

PR #420, opened 2021-01-11
Fix crate publishing

### LINKED ISSUE #412: Cargo publish fails on CI
The failure happens as the task is using Alpine.

[log.zip](https://github.com/mozilla/rust-code-analysis/files/5787976/log.zip)

Luni-4:
This PR fixes #412

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -321,7 +321,7 @@ tasks:
             features:
               taskclusterProxy: true
             maxRunTime: 3600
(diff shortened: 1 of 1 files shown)


marco-c (approved):

marco-c:
I forgot, we also need to run `cargo package` as part of the Linux release build task, to ensure the package works as expected (this way using `--no-verify` for publish actually has no real effect).

PR #421, opened 2021-01-11
Do not strip binaries

Luni-4:
`Rust` programs are able to print a stacktrace thanks to the debug info in the file, so stripping them will probably kill this nice and very useful feature. It would a way better in general to propose the binary as compiled by `rustc`

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -241,7 +241,6 @@ tasks:
                  pushd rust-code-analysis-cli && cargo package --all-features --target x86_64-unknown-linux-musl && popd &&
                  pushd rust-code-analysis-web && cargo package --all-features --target x86_64-unknown-linux-musl && popd &&
                  cd target/x86_64-unknown-linux-musl/release &&
-                 strip rust-code-analysis-cli rust-code-analysis-web &&
                  tar -zvcf /home/rust/rust-code-analysis-linux-cli-x86_64.tar.gz rust-code-analysis-cli &&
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -307,7 +306,7 @@ tasks:
 
       - $if: 'tasks_for == "github-push" && head_branch[:10] == "refs/tags/"'
         then:
-          taskId: {$eval: as_slugid("strip_windows_binary")}
```
marco-c, line 310: We could drop this task now and just generate the zipin the `build_windows_release` task.
If you don't want to do it now, please file an issue and add it to the CI project (see, nothing is ever complete :P).

marco-c (approved):

.taskcluster.yml
Luni-4: > (see, nothing is ever complete :P).

(for now :P)

Luni-4: #430

marco-c: #430 is about reducing the size of the binary, we should also file an issue about dropping the `strip` task (which is no longer needed, and might not be required for #430 either).

Luni-4: It could be required for #430, since it will be much simpler to install software that might reduce the size of the binaries in a Linux image instead of a Windows one.

marco-c: All the suggestions from the links in #430 refer to things to change which are either in the Rust source code or in the Cargo.toml and don't require external tools.

Luni-4: You're right, updated the issue, forgot a link :P

PR #424, opened 2021-01-11
Update continuous integration

Luni-4:
- Replace the deprecated `all` option with the `workspace` one
- Update `task-boot` version
- Reduce an unreadable long line into a shorter one


marco-c:
> Replace the deprecated all option with the workspace one
> Update task-boot version
> Reduce an unreadable long line into a shorter one

Right now you have one commit for the first two changes, and one commit for the last change.
Could you split the changes in three commits instead?

marco-c (approved):

PR #427, opened 2021-01-13
Remove an unused unsafe function

Luni-4:
This function has never been used internally and furthermore contains unsafe code

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -58,17 +58,6 @@ macro_rules! mk_impl_lang {
     ( $( ($camel:ident, $name:ident, $display: expr) ),* ) => {
         impl LANG {
 
-            #[allow(dead_code)]
-            pub(crate) fn get_language(&self) -> Language {
-                unsafe {
-                    match self {
-                        $(
-                            LANG::$camel => $name(),
-                        )*
-                    }
-                }
-            }
-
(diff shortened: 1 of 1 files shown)


src/macros.rs
```diff
@@ -58,17 +58,6 @@ macro_rules! mk_impl_lang {
     ( $( ($camel:ident, $name:ident, $display: expr) ),* ) => {
         impl LANG {
 
-            #[allow(dead_code)]
```
marco-c, line 61: It seems this function was dead code on purpose. @calixteman do you remember why?

Luni-4: I have added the dead code to fix documentation. https://github.com/mozilla/rust-code-analysis/commit/c5b9bf7ba6f7ae3f4b59d8becf7a3fe5a781acb9#diff-315c02cd05738da173861537577d159833f70f79cfda8cd7cf1a0d7a28ace31bR48

marco-c: Is it not needed anymore to fix documentation?

Luni-4: If I remove the function, the problem does not pose anymore. And it is effectively dead code, so we can remove it. It is an internal function also, so no problem for users

marco-c (approved):

PR #428, opened 2021-01-13
Remove a useless macro

Luni-4:
This macro can be removed and its code can be inserted in the relative trait's method.


Luni-4:
Depends no #427

marco-c (approved):

Issue #442, opened 2021-02-04
Replace tree-sitter submodules with Rust bindings


Luni-4:
Some of the `tree-sitter` grammars can be used as `Cargo.toml` dependencies. In this way, some submodules can be removed.

Pros
- A cleaner repository
- Avoid breaks during the update of a tree-sitter grammar due to some problems in handling submodules
- Less options and commands to clone the project

Cons 
- Impossibility to change the grammars directly in this repo, the grammars have to be modified upstream

Submodules languages:

- [x] tree-sitter-java (#449)
- [x] tree-sitter-python (#464)
- [x] tree-sitter-cpp (#498)
- [x] tree-sitter-rust (#594)
- [x] tree-sitter-javascript (#463)
- [x] tree-sitter-typescript (#592)

marco-c:
I've removed it from the "API & User Interface" project since this is not affecting end users of the project, but only people who want to build it.

PR #449, opened 2021-02-08
Delete Java submodule

Luni-4:
This PR removes the `tree-sitter-java` submodule, fixing the first item of #442

The `enums` submodule has been updated accordingly. 

The sequence of commands to test whether the keywords are changed, they haven't produced any difference:
```
cargo clean
cargo run -- -lrust -o ../src/languages
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ serde = { version = "^1.0", features = ["derive"] }
 termcolor = "^1.1"
 
 tree-sitter = "^0.17"
-tree-sitter-java = "0.16"
+tree-sitter-java = "^0.16"
 
 [dev-dependencies]
 pretty_assertions = "^0.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,9 @@ phf = { version = "^0.8", features = ["macros"] }
 regex = "^1.4"
 serde = { version = "^1.0", features = ["derive"] }
 termcolor = "^1.1"
+
 tree-sitter = "^0.17"
+tree-sitter-java = "0.16"
 
 [dev-dependencies]
 pretty_assertions = "^0.6"
--- a/enums/Cargo.toml
+++ b/enums/Cargo.toml
@@ -17,4 +17,4 @@ phf_codegen = "^0.8"
 libc = "^0.2"
 
 tree-sitter = "^0.17"
-tree-sitter-java = "0.16"
+tree-sitter-java = "^0.16"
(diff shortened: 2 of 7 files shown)


enums/Cargo.toml
```diff
@@ -10,9 +10,11 @@ cc = "^1.0"
 phf_codegen = "^0.8"
 
 [dependencies]
-tree-sitter = "^0.16"
```
marco-c, line 13: Could you update the tree-sitter dependency in a separate PR (which I will land first)?

enums/src/macros.rs
```diff
@@ -23,10 +14,15 @@ macro_rules! mk_enum {
 macro_rules! mk_get_language {
     ( $( ($camel:ident, $name:ident) ),* ) => {
         pub fn get_language(lang: &LANG) -> Language {
-            unsafe {
+              if let LANG::Java = lang {
```
marco-c, line 17: It's a pain that we have to specify all languages manually now :(

Luni-4: The other solution would be splitting up the big macro in `enums/src/languages.rs`. One macro for bindings and one for directories.

enums/Cargo.toml
Luni-4: Done in #451

enums/src/macros.rs
Luni-4: @marco-c 

Thinking about this one in depth, this is just a temporary solution, because at the end only the Mozilla languages will be to need specified manually.

Luni-4: If we decide to add bindings to the Mozilla languages too and use the `crate_name = { path ="..." }` notation, the structure will return the same as it is now.

Cargo.toml
```diff
@@ -29,7 +29,9 @@ phf = { version = "^0.8", features = ["macros"] }
 regex = "^1.4"
 serde = { version = "^1.0", features = ["derive"] }
 termcolor = "^1.1"
+
 tree-sitter = "^0.17"
+tree-sitter-java = "0.16"
```
marco-c, line 34: ^0.16, so we accept patch updates

enums/Cargo.toml
```diff
@@ -17,3 +17,4 @@ phf_codegen = "^0.8"
 libc = "^0.2"
 
 tree-sitter = "^0.17"
+tree-sitter-java = "0.16"
```
marco-c, line 20: Same here

enums/src/macros.rs
marco-c: That sounds good to me

marco-c (approved):

PR #452, opened 2021-02-08
Consider enums dependencies for dependabot

### LINKED ISSUE #450: Dependabot is not considering dependencies from enums/Cargo.toml
It is monitoring Cargo.toml, Cargo.lock, rust-code-analysis-cli/Cargo.toml, rust-code-analysis-web/Cargo.toml, but not enums/Cargo.toml:
immagine

Luni-4:
This PR fixes #450

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,11 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 99
+- package-ecosystem: cargo
+  directory: "/enums"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 99
 - package-ecosystem: gitsubmodule
   directory: "/"
   schedule:


.github/dependabot.yml
```diff
@@ -10,3 +10,8 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 99
+- package-ecosystem: enums
```
marco-c, line 13: It should be `cargo`, no?

.github/dependabot.yml
```diff
@@ -10,3 +10,8 @@ updates:
...
+  directory: "/enums"
```
marco-c, line 14: Why do we need to specify it? We don't specify rust-code-analysis-cli and rust-code-analysis-web, but they are picked up by Dependabot anyways

Luni-4: Because we have excluded it: https://github.com/mozilla/rust-code-analysis/blob/master/Cargo.toml#L39

marco-c (approved):

PR #459, opened 2021-02-08
Test the enums workspace on CI

### LINKED ISSUE #457: Add a CI check for the enums workspace
Currently we don't check whether the `enums` workspace works as expected among the various dependabot updates, so we need to add a task to our CI.

Luni-4:
This PR fixes #457

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -46,7 +46,9 @@ tasks:
                git -c advice.detachedHead=false checkout --recurse-submodules ${head_rev} &&
                pip3 install --quiet pre-commit &&
                pre-commit run -a --show-diff-on-failure &&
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -53,6 +53,28 @@ tasks:
           owner: cdenizet@mozilla.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+      - taskId: {$eval: as_slugid("test_enums_workspace")}
```
marco-c, line 56: Let's add it to the `lint_test_task` task instead, so we don't pay the setup price twice.

marco-c (approved):

PR #463, opened 2021-02-09
Delete JavaScript submodule

Luni-4:
This PR removes the `tree-sitter-javascript` submodule, fixing the fifth item of #442

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,12 +7,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "78d1833b3838dbe990df0f1f87baf640cf6146e898166afe401839d1b001e570"
 dependencies = [
  "bitflags",
- "bytes",
+ "bytes 0.5.6",
  "futures-core",
  "futures-sink",
  "log",
- "pin-project 0.4.23",
- "tokio 0.2.22",
+ "pin-project 0.4.28",
+ "tokio 0.2.25",
  "tokio-util",
 ]
 
@@ -50,7 +50,7 @@ dependencies = [
  "base64",
  "bitflags",
  "brotli2",
- "bytes",
+ "bytes 0.5.6",
  "cookie",
  "copyless",
  "derive_more",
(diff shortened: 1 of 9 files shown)


calixteman:
This change will break update for moz-js grammar:
https://github.com/mozilla/rust-code-analysis/blob/master/tree-sitter-mozjs/grammar.js#L1

Luni-4:
Thanks for the catch @calixteman! 

Is it possible to use directly https://github.com/tree-sitter/tree-sitter-javascript/blob/master/grammar.js?

Luni-4:
The simplest solution would be copying that file inside the `moz-js` language, but perhaps there is a better alternative

marco-c (approved):

calixteman:
Several possibilities:
 - use docker to build
 - update tree-sitter tool to handle the case where a dependency is a remote one

Anyway at some point we'll need to do something to build those grammars (js, cpp, ...) and so we'll probably need to use docker or whatever.

marco-c:
> This change will break update for moz-js grammar:

Oh, I missed that. We should catch that in CI!

marco-c:
> The simplest solution would be copying that file inside the moz-js language, but perhaps there is a better alternative

That is not a good solution. With the submodule, we automatically get moz-js updated when we update js. After this patch, that update would be manual.

Luni-4:
No more blocked by `mozjs` grammar dependency, but from the fact that our submodule version is more recent than the one from the current `tree-sitter-javascript` crate

Cargo.lock
```diff
@@ -2154,29 +2108,40 @@ dependencies = [
...
+checksum = "97d095ae15e245a057c8e8451bab9b3ee1e1f68e9ba2b4fbc18d0ac5237835f2"
+dependencies = [
+ "pin-project 1.0.6",
+ "tracing",
+]
+
 [[package]]
 name = "tree-sitter"
-version = "0.19.2"
+version = "0.19.3"
```
marco-c, line 2142: We shouldn't update to 0.19.3 without checking the metrics diffs

Luni-4:
@marco-c 

No differences in metrics for `tree-sitter v0.19.3` https://community-tc.services.mozilla.com/api/queue/v1/task/Fn4SPov-Si-R8CyvewED6A/runs/0/artifacts/public%2Fjson-diffs-and-minimal-tests.tar.gz

marco-c (approved):

PR #466, opened 2021-02-09
Bump tree-sitter-python

Luni-4:
This PR updates the `tree-sitter-python` submdoule to `v0.17`.


marco-c:
@Luni-4 can we close this in favor of #512?

Issue #469, opened 2021-02-11
Write some scripts to generate mozcpp and mozjs grammars automatically


Luni-4:
It would be helpful to create two scripts that generate automatically the `mozjs` and `mozcpp` grammars, because doing that manually is a bit cumbersome and prone to errors

- [x] mozcpp (#470)
- [x] mozjs (#471)

PR #470, opened 2021-02-11
Generate mozcpp grammar automatically

Luni-4:
This PR fixes the first item of #469

### DIFF
--- a/generate-moz-grammars/generate-mozcpp.sh
+++ b/generate-moz-grammars/generate-mozcpp.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+
+# This script updates the mozcpp grammar automatically.
+#
+# Usage: ./generate-moz-grammars/generate-mozcpp.sh
+
+# FIXME we need to remove this line once we are going to use
+# the tree-sitter-cpp bindings
+# Get the tree-sitter-cpp submodule version
+TS_CPP_VERSION=`git submodule status tree-sitter-cpp | awk '{ print $1 }'`
+
+# Enter the mozcpp directory
+pushd tree-sitter-mozcpp
+
+# Create tree-sitter-cpp directory
+mkdir -p tree-sitter-cpp
+
+# Enter tree-sitter-cpp directory
+pushd tree-sitter-cpp
+
+# Shallow clone tree-sitter-cpp to a specific revision
+git init
+git remote add origin https://github.com/tree-sitter/tree-sitter-cpp.git
+git fetch --depth 1 origin $TS_CPP_VERSION
+git checkout FETCH_HEAD
+
+# Install tree-sitter-cpp dependencies
+npm install -y
+
+# Exit tree-sitter-cpp directory
+popd
+
+# Init npm
+npm init -y
+
+# Install a small module that lets the parser be used from Node
+npm install --save nan
+
+# Install the Tree-sitter CLI
+npm install --save-dev tree-sitter-cli
+
+# Generate moz-cpp grammar
+./node_modules/.bin/tree-sitter generate
+
+# Delete node_modules
+rm -rf node_modules
+
+# Delete package files
+rm -rf package-lock.json package.json
+
+# Delete tree-sitter-cpp directory
+rm -rf tree-sitter-cpp
+
+# Exit tree-sitter-mozcpp directory
+popd
+
+# Enter enums directory
+pushd enums
+
+# Recreate the grammar for rust-code-analysis
+cargo clean && cargo run -- -lrust -o ../src/languages
+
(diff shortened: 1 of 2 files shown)


generate-moz-grammars/generate-mozcpp.sh
```diff
@@ -0,0 +1,71 @@
...
+
+# This script updates the mozcpp grammar automatically.
+#
+# Usage: ./generate-moz-grammars/generate-mozcpp.sh
+
+# FIXME we need to remove this line once we are going to use
+# the tree-sitter-cpp bindings
+# Update tree-sitter-cpp submodule such that is the same one used within the
+# tree-sitter-mozcpp directory
+git submodule update --remote tree-sitter-cpp
```
marco-c, line 11: Instead of updating the tree-sitter-cpp submodule as part of this, we can just get its revision and, in the tree-sitter-mozcpp directory, clone the repo at that revision

generate-moz-grammars/generate-mozcpp.sh
```diff
@@ -0,0 +1,71 @@
...
+./node_modules/.bin/tree-sitter generate
+
+# Delete node_modules
+rm -rf node_modules
+
+# Delete package files
+rm -rf package-lock.json package.json
+
+# Delete tree-sitter-cpp
+rm -rf tree-sitter-cpp
```
marco-c, line 52: We should remove the `tree-sitter-c` directory too, right?

Luni-4: Since `tree-sitter-c` is inside `tree-sitter-cpp`, we remove both of them in one fell swoop

generate-moz-grammars/generate-mozcpp.sh
Luni-4: Good idea! :smiley:

generate-moz-grammars/generate-mozcpp.sh
```diff
@@ -0,0 +1,73 @@
...
+# FIXME we need to remove this line once we are going to use
+# the tree-sitter-cpp bindings
+# Get the tree-sitter-cpp submodule version
+TS_CPP_VERSION=`git submodule status tree-sitter-cpp | awk '{ print $1 }'`
+
+# Enter the mozcpp directory
+pushd tree-sitter-mozcpp
+
+# Download tree-sitter-cpp
+git clone https://github.com/tree-sitter/tree-sitter-cpp.git
```
marco-c, line 16: Nit: it'd be nice to use shallow clones so we directly clone the right revision and only the files at that revision

generate-moz-grammars/generate-mozcpp.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Enter tree-sitter-cpp directory
+pushd tree-sitter-cpp
+
+# Checkout the tree-sitter-cpp submodule version
+git checkout $TS_CPP_VERSION
+
+# Clone tree-sitter-c repository
+git clone https://github.com/tree-sitter/tree-sitter-c
```
marco-c, line 25: Don't we need to set this to the specific revision tree-sitter-cpp expects?

generate-moz-grammars/generate-mozcpp.sh
```diff
@@ -0,0 +1,76 @@
...
+pushd tree-sitter-cpp
+
+# Shallow clone tree-sitter-cpp to a specific revision
+git init
+git remote add origin https://github.com/tree-sitter/tree-sitter-cpp.git
+git fetch --depth 1 origin $TS_CPP_VERSION
+git checkout FETCH_HEAD
+
+# Clone tree-sitter-c repository
+git clone --depth 1 --branch 0.16.1 https://github.com/tree-sitter/tree-sitter-c.git
```
marco-c, line 28: Why 0.16.1?

Luni-4: Because it depends on https://github.com/tree-sitter/tree-sitter-cpp/blob/master/package.json#L16

marco-c: Instead of cloning the repo, why don't we use `npm` to install the tree-sitter-cpp dependencies? Then the path to tree-sitter-c will just be a subdirectory of node_modules

1. Clone tree-sitter-cpp at the right revision
2. cd tree-sitter-cpp
3. npm install
4. Use the tree-sitter-c path from node_modules

Luni-4: Tried your method, but 4. doesn't exist. I can't find any path in `node_modules`

marco-c (approved):

PR #477, opened 2021-02-18
Divide minimal tests per metric

Luni-4:
Instead of having a big directory containing all minimal tests, it is better to split them up per metric

### DIFF
--- a/check-submodules.sh
+++ b/check-submodules.sh
@@ -45,5 +45,31 @@ ls /tmp/$SUBMODULE_NAME-new | wc -l
 # Create artifacts to be uploaded (if there are any)
 COMPARE=/tmp/$SUBMODULE_NAME-compare
 if [ "$(ls -A $COMPARE)" ]; then
-    tar -czvf /tmp/json-diffs-and-minimal-tests.tar.gz $COMPARE
+    # Maximum number of considered minimal tests for a metric
+    MT_THRESHOLD=30
+
+    # Array containing the considered metrics
+    # TODO: Implement a command into rust-code-analysis-cli that returns all
+    # computed metrics https://github.com/mozilla/rust-code-analysis/issues/478
+    METRICS=("cognitive" "sloc" "ploc" "lloc" "cloc" "blank" "cyclomatic" "halstead" "nom" "nexits" "nargs")
+
+    # Output directory name
+    OUTPUT_DIR=/tmp/output-$SUBMODULE_NAME
+
+    # Create output directory
+    mkdir -p $OUTPUT_DIR
+
+    # Retrieve minimal tests for a metric
+    for METRIC in "${METRICS[@]}"
+    do
+
+        FILES=`grep -i -l $METRIC $COMPARE/* | head -$MT_THRESHOLD`
+        if [ ${#FILES[@]} -ne 0 ]
+        then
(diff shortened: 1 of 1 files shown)


marco-c:
Let's output both the full minimal tests file, and a reduced one with fewer tests. So if somebody wants to look at more than what we decided (30), they can.

check-submodules.sh
```diff
@@ -45,5 +45,29 @@ ls /tmp/$SUBMODULE_NAME-new | wc -l
 # Create artifacts to be uploaded (if there are any)
 COMPARE=/tmp/$SUBMODULE_NAME-compare
 if [ "$(ls -A $COMPARE)" ]; then
-    tar -czvf /tmp/json-diffs-and-minimal-tests.tar.gz $COMPARE
+    # Maximum number of considered minimal tests for a metric
+    MT_THRESHOLD=30
+
+    # Array containing the considered metrics
+    METRICS=("cognitive" "sloc" "ploc" "lloc" "cloc" "blank" "cyclomatic" "halstead" "nom" "nexits" "nargs")
```
marco-c, line 52: This is a bit brittle. We should have a single place where we define all the metrics.
We could add a `--list-metrics` command to rca to list all the supported metrics and use it here.

Luni-4: I agree, I would add a TODO for now and leave as-is

Luni-4: @marco-c 

Don't you think that that type of command is useless except for this specific goal? Don't know if it's worth implementing it

marco-c: It'd be nice to have such a command anyway (one to list the metrics, one to list the metrics together with a short explanation about them).
It could be useful in other cases where you need to know all the available metrics.

marco-c: If you don't want to do it now, please file an issue and add a TODO as a comment here pointing to the issue

Luni-4: #478

check-submodules.sh
```diff
@@ -45,5 +45,31 @@ ls /tmp/$SUBMODULE_NAME-new | wc -l
...
+
+    # Retrieve minimal tests for a metric
+    for METRIC in "${METRICS[@]}"
+    do
+
+        FILES=`grep -i -l $METRIC $COMPARE/* | head -$MT_THRESHOLD`
+        if [ ${#FILES[@]} -ne 0 ]
+        then
+            mkdir -p $OUTPUT_DIR/$METRIC
+            mv $FILES $OUTPUT_DIR/$METRIC
```
marco-c, line 70: Some files could belong to more than one metric, let's cp instead of mv

Luni-4: I added the `mv` command to reduce the final filesize. Having lots of files, we may *assume* that we would have *at least* 30 files per each metric. In addition, with the `cp` we would have the same files in each metric-directory. I don't think it won't be a problem not varying the files though.

marco-c: The filesize will be relatively small anyway, as compression should work very well when you have duplicated files

marco-c (approved):

check-submodules.sh
Luni-4: Sorry, I was talking about the uncompressed filesize

PR #481, opened 2021-02-19
Update json-minimal-tests crate

Luni-4:
This PR updates `json-minimal-tests` crate such that non-utf-8 files are ignored


marco-c:
Instead of ignoring non-utf-8 files, why don't you ignore decoding errors?

marco-c (approved):

PR #498, opened 2021-03-03
Remove tree-sitter-cpp submodule

Luni-4:
This PR removes the `tree-sitter-submodule` in this way:

1. Since the grammar for the cpp language is completely managed by `tree-sitter-mozcpp`, first of all we need to make `tree-sitter-mozcpp` completely independent from the `tree-sitter-cpp` submodule. To do so, we make a copy of the `scanner.cc` file during the generation of the grammar

2. Now we can remove the `tree-sitter-cpp` submodule

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,11 @@ updates:
     interval: weekly
   open-pull-requests-limit: 99
 - package-ecosystem: cargo
+  directory: "/tree-sitter-mozcpp"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 99
+- package-ecosystem: cargo
   directory: "/enums"
   schedule:
     interval: weekly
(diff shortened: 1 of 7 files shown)


marco-c (approved):
I wonder if there isn't a better way to reuse a grammar, extending it... For now, this is OK.

marco-c:
Clippy is failing though, please fix that before merging

Luni-4:
> I wonder if there isn't a better way to reuse a grammar, extending it... For now, this is OK.

We could extend it such as the `tree-sitter-mozjs` grammar. But perhaps there is a specific reason why the same method adopted for `tree-sitter-mozjs` haven't been used. @calixteman do you remember why?

Luni-4:
@marco-c 

Now it is fixed

marco-c (approved):

PR #512, opened 2021-03-08
Update Python grammar to v0.17

Luni-4:
This PR adds more metrics tests for `Python` in order to update the grammar to `v0.17`


Luni-4:
@marco-c 

Can't find any other minimal tests different one from the another, so this PR is now ready to be reviewed

src/metrics/loc.rs
```diff
@@ -541,6 +543,25 @@ mod tests {
...
+        // More lines of the same instruction were counted as blank lines
+        check_metrics!(
+            "capabilities[\"goog:chromeOptions\"][\"androidPackage\"] = \\
+                \"org.chromium.weblayer.shell\"",
+            "foo.py",
+            PythonParser,
+            loc,
+            [
+                (sloc, 2, usize),
+                (ploc, 2, usize),
```
marco-c, line 557: Should ploc be 1 too?

src/metrics/cognitive.rs
```diff
@@ -477,6 +484,54 @@ mod tests {
...
+
+    #[test]
+    fn python_elif_function() {
+        // Boolean expressions containing `And` and `Or` operators were not
+        // considered in `elif` statements
+        check_metrics!(
+            "def f(a, b):
+                if a and b:  # +2 (+1 and)
+                   return 1
+                elif c and d: # +2 (+1 and)
```
marco-c, line 525: Can you also add an additional test with two `elif`s?

marco-c (dismissed):

src/metrics/loc.rs
Luni-4: No, `ploc` is correct because we have two physical lines and one single instruction (so `lloc` is 1 in this case)

src/metrics/cognitive.rs
Luni-4: Sure! :)

marco-c (approved):

PR #529, opened 2021-03-10
Bump tree-sitter-ccomment to v0.17

Luni-4:
This PR bumps `tree-sitter-ccomment` to `v0.17`. Since the `grammar.js` file hasn't been modified, there are no changes in our language.


marco-c:
> Since the grammar.js file hasn't been modified, there are no changes in our language.

It actually has been slightly modified, but I guess that change didn't have any effect?

Luni-4:
> > Since the grammar.js file hasn't been modified, there are no changes in our language.
> 
> It actually has been slightly modified, but I guess that change didn't have any effect?

Yeah, exactly

PR #538, opened 2021-03-10
Do not remove package files from moz grammars

Luni-4:
When a grammar is updated, package files need to be maintained in order to pin up dependencies so that it is impossible to have indefinite `tree-sitter-cli` versions


marco-c:
Merged as part of #539.

marco-c:
Oops I clicked on merge, this created an empty commit: 08a5bbc51fd87f6265063e4e22e7f7ec15b8a10b...

PR #543, opened 2021-03-11
Check metrics when a tree-sitter crate is updated

Luni-4:
This PR checks whether there are metric changes when a `tree-sitter` crate of a determined grammar, or the crate containing the `tree-sitter` library, has been updated. The metrics-checker is the same one used for submodules

### DIFF
--- a/check-tree-sitter-crates.sh
+++ b/check-tree-sitter-crates.sh
@@ -1,0 +1,107 @@
+#!/bin/bash
+
+# Stop at the first error
+set -e
+
+# Get all tree-sitter crates from the analyzed branch Cargo.toml
+TS_CRATES=`grep "tree-sitter-*" Cargo.toml | tr -d ' '`
+
+# Disable/Enable CI flag
+RUN_CI="no"
+
+# Temporary master branch Cargo.toml filename
+MASTER_CARGO_TOML="master-cargo.toml"
+
+# Download master branch Cargo.toml and save it in a temporary file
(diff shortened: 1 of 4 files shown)


check-tree-sitter-crates.sh
```diff
@@ -0,0 +1,107 @@
...
+TS_CRATES=`grep "tree-sitter-*" Cargo.toml | tr -d ' '`
+
+# Disable/Enable CI flag
+RUN_CI="no"
+
+# Temporary master branch Cargo.toml filename
+MASTER_CARGO_TOML="master-cargo.toml"
+
+# Download master branch Cargo.toml and save it in a temporary file
+wget -LqO - https://raw.githubusercontent.com/mozilla/rust-code-analysis/master/Cargo.toml | tr -d ' ' > $MASTER_CARGO_TOML
```
marco-c, line 16: We could use git to get the content of the Cargo.toml file before the PR.

marco-c: Note: this won't work if we start using branches (as we'd always compare to master, we might end up running the checks always in the branch)

check-tree-sitter-crates.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
+
+# Stop at the first error
+set -e
+
+# Get all tree-sitter crates from the analyzed branch Cargo.toml
+TS_CRATES=`grep "tree-sitter-*" Cargo.toml | tr -d ' '`
```
marco-c, line 7: This won't include the `tree-sitter` dependency, right?

Luni-4: Yes, it includes the `tree-sitter` dependency

marco-c (approved):

PR #545, opened 2021-03-12
Allow finding all nodes and nodes containing a string

abrown:
When using the `-f` option in `rust-code-analysis-cli` to find nodes, the current behavior is to return all found nodes if the passed filter string is not one of `call|comment|error|string|function|<u16>`. I found it useful to be able to find nodes that contained a certain string (e.g. `expr`) so I added this functionality. Because in testing I found that an empty `-f` never creates the "accept all" filter, I added a keyword `"all"` to enable this behavior (also quite useful). The filters now accept `all|call|comment|error|string|function|<u16>|<string>`.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -121,6 +121,7 @@ impl<T: 'static + TSLanguage + Checker + Getter + Alterator + CodeMetricsT> Pars
         for f in filters.iter() {
             let f = f.as_str();
             match f {
+                "all" => res.push(Box::new(|_: &Node| -> bool { true })),
                 "call" => res.push(Box::new(T::is_call)),
                 "comment" => res.push(Box::new(T::is_comment)),
                 "error" => res.push(Box::new(T::is_error)),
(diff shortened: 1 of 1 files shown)


calixteman (approved):
LGTM.
Thank you for your patch.

calixteman:
My first goal with this option was mainly to find errors in asts to fix `tree-sitter` grammars.
I'd like to implement a kind of `grep` where we could match some string in the context of some nodes.
What's your use case ?

abrown:
Right now I'm just using `rust-code-analysis-cli` to do some cross-language exploring (BTW, great tool!). I think it would cool to eventually be able to search for tree patterns (e.g. "while loops that contain..."); maybe that's what you meant by "strings in the context of some nodes" and `grep`. I would assume that I would have to switch over to using the library instead of the CLI but that is fine. Some preliminary questions along these lines: 
 - Is there a way to ask the tool to find X and then look inside of X for Y?
 - And is there a way to ask the tool once we've found an identifier Z in expression X, find me all uses/defines of Z outside of X?

marco-c:
@abrown we have thought about something along those lines. Our dream would be to have something like https://www.youtube.com/watch?v=yqi8U8Q0h2g, but cross-language.

marco-c:
Sorry, I forgot to actually answer your questions.

> Is there a way to ask the tool to find X and then look inside of X for Y?

I don't think so, but you could use the `--dump` option to output a tree, and then implement something like that on top of the output.

> And is there a way to ask the tool once we've found an identifier Z in expression X, find me all uses/defines of Z outside of X?

Nope, I guess you could build it on top of the `--dump` option too.

@calixteman might have more ideas, as he's way more expert than me on this project.

PR #547, opened 2021-03-12
Retrieve Cargo.toml from master branch using git

Luni-4:
It is not necessary to download `Cargo.toml` from master branch, it can be retrieved directly through `git`


marco-c:
We should get the contents from the last commit before the PR. If the PR's target is master, that's the latest commit on master, but the PR's target could be another branch.

Luni-4:
> We should get the contents from the last commit before the PR. If the PR's target is master, that's the latest commit on master, but the PR's target could be another branch.

Done, thanks!

marco-c:
Unfortunately this won't work because the "master" branch is the branch from the fork. We'll handle this in the future (filed #548 for it).

PR #560, opened 2021-03-17
Print only the files that are in a directory but not in the other one

Luni-4:
This PR prints on `stdout` only the files that are in a directory but not in the other one

### DIFF
--- a/check-submodules.sh
+++ b/check-submodules.sh
@@ -44,10 +44,11 @@ echo "$SUBMODULE_NAME-old: $OLD"
 echo "$SUBMODULE_NAME-new: $NEW"
 
 # If metrics directories differ in number of files,
-# print the difference of files
+# print only the files that are in a directory but not in the other one
 if [ $OLD != $NEW ]
 then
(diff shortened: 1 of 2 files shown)


check-submodules.sh
```diff
@@ -44,10 +44,11 @@ echo "$SUBMODULE_NAME-old: $OLD"
 echo "$SUBMODULE_NAME-new: $NEW"
 
 # If metrics directories differ in number of files,
-# print the difference of files
+# print only the files that are in a directory but not in the other one
 if [ $OLD != $NEW ]
 then
-    diff -rq /tmp/$SUBMODULE_NAME-old /tmp/$SUBMODULE_NAME-new
+    ONLY_FILES=`diff -q /tmp/$SUBMODULE_NAME-old /tmp/$SUBMODULE_NAME-new | grep "Only in"`
```
marco-c, line 50: Don't we still need the `-r` option?

Luni-4: We can remove the `-r` option since the recursion are not necessary for the directories we are scanning through. They are flat

marco-c (approved):

PR #563, opened 2021-03-18
Improve the part of CI related to group by minimal tests

Luni-4:
This PR improves the part of CI related to group by minimal tests:

- Checks if a shell variable is full instead of non-empty
-  Add a `.` before a metric name in order to find only the minimal tests which effectively contain a change for that metric. In this way, false positives, caused by the presence of a metric name inside source codes, are avoided.

### DIFF
--- a/check-submodules.sh
+++ b/check-submodules.sh
@@ -75,7 +75,8 @@ if [ "$(ls -A $COMPARE)" ]; then
     for METRIC in "${METRICS[@]}"
     do
 
-        FILES=`grep -r -i -l $METRIC $COMPARE | head -$MT_THRESHOLD`
+        PREFIX_METRIC="\.$METRIC"
+        FILES=`grep -r -i -l $PREFIX_METRIC $COMPARE | head -$MT_THRESHOLD`
         if [ -n "$FILES" ]
         then
             mkdir -p $OUTPUT_DIR/$METRIC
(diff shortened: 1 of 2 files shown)


marco-c:
> That was not precise because lots of false positives have been found

Do you know why? This change is a bit suboptimal because we have to keep the list of metrics up to date in several places

Luni-4:
> > That was not precise because lots of false positives have been found
> 
> Do you know why? This change is a bit suboptimal because we have to keep the list of metrics up to date in several places

Yes, for example, the `blank` metric. Using that as keyword we get `_blank`, `Blank` present in source codes. So we get some files in the `blank` directory that doesn't have the `blank` metric changed at all.
To avoid this, it's better to use parts of the paths provided by `json-minimal-tests`. Later I could add a cli option to `json-minimal-tests` to return all metrics subpath automatically.

marco-c:
What if you just make json-minimal-tests write METRICNAME___ in the file name, so you can be pretty confident that if you look e.g. for "blank___" you'll only find the actual files you want

Luni-4:
> What if you just make json-minimal-tests write METRICNAME___ in the file name, so you can be pretty confident that if you look e.g. for "blank___" you'll only find the actual files you want

IIRC there were some m-c files with that structure, so it's better a path to avoid false positives

marco-c:
>IIRC there were some m-c files with that structure, so it's better a path to avoid false positives

You can do all sort of crazy text like `blank_SURELYUNIQUE_` or `blank_jsonminimaltests_` or whatever

Luni-4:
@marco-c 

Fixed the problem and updated the description of this PR

marco-c (approved):

PR #580, opened 2021-03-25
Fix a corner case for blank metric

Luni-4:
This PR fixes blank metric when there are code lines and comments on the same line. Before of this PR, those kinds of lines were counted twice.

### DIFF
--- a/src/metrics/loc.rs
+++ b/src/metrics/loc.rs
@@ -14,7 +14,9 @@ pub struct Stats {
     unit: bool,
     lines: FxHashSet<usize>,
     logical_lines: usize,
-    comment_lines: usize,
+    only_comment_lines: usize,
+    code_comment_lines: usize,
+    comment_line_end: Option<usize>,
 }
 
 impl Serialize for Stats {
@@ -58,7 +60,8 @@ impl Stats {
         self.logical_lines += other.logical_lines;
 
         // Merge cloc lines
-        self.comment_lines += other.comment_lines;
+        self.only_comment_lines += other.only_comment_lines;
+        self.code_comment_lines += other.code_comment_lines;
     }
 
     /// The `Sloc` metric.
@@ -104,7 +107,7 @@ impl Stats {
     pub fn cloc(&self) -> f64 {
         // Comments are counted regardless of their placement
         // https://en.wikipedia.org/wiki/Source_lines_of_code
-        self.comment_lines as f64
+        (self.only_comment_lines + self.code_comment_lines) as f64
     }
 
     /// The `Blank` metric.
@@ -112,12 +115,7 @@ impl Stats {
     /// Counts the number of blank lines in a scope
     #[inline(always)]
     pub fn blank(&self) -> f64 {
-        // This metric counts the number of blank lines in a code
-        // The if construct is needed because sometimes lloc and cloc
-        // coincide on the same lines, in that case lloc + cloc could be greater
-        // than the number of lines of a file.
-        let blank = self.sloc() - self.ploc() - self.cloc();
-        blank.max(0.0)
+        self.sloc() - self.ploc() - self.only_comment_lines as f64
     }
 }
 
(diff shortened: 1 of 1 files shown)


marco-c (changes requested):
Your test is a lucky case where this works, but it doesn't work in general. I think you actually have to subtract lines that are both ploc and cloc only once.
For example, try removing the blank line in python_negative_blank (the result is "blank == 2", while it should be 0). Or, add another comment at the end of another source line (again, the result is "blank == 2", while it should still be 1).

Luni-4:
> Your test is a lucky case where this works, but it doesn't work in general. I think you actually have to subtract lines that are both ploc and cloc only once.
> For example, try removing the blank line in python_negative_blank (the result is "blank == 2", while it should be 0). Or, add another comment at the end of another source line (again, the result is "blank == 2", while it should still be 1).

It should be fixed now

src/metrics/loc.rs
```diff
@@ -457,6 +468,109 @@ mod tests {
         );
     }
 
+    #[test]
+    fn python_negative_blank() {
+        check_metrics!(
+            "def ConnectToUpdateServer():
```
marco-c, line 474: Can you also add the other two test cases I mentioned in the earlier review?

Luni-4: Sure! :)

src/metrics/loc.rs
```diff
@@ -142,6 +139,20 @@ fn init(node: &Node, stats: &mut Stats, is_func_space: bool, is_unit: bool) -> (
     (start, end)
 }
 
+#[inline(always)]
+// This function discriminates among the comments that are next to code lines and
+// the ones that are on a single line. This difference is necessary in
+// order to avoid having a wrong count for the blank metric.
+fn cloc_lines(stats: &mut Stats, start: usize, end: usize) {
```
marco-c, line 146:
```suggestion
fn add_cloc_lines(stats: &mut Stats, start: usize, end: usize) {
```

src/metrics/loc.rs
```diff
@@ -142,6 +139,20 @@ fn init(node: &Node, stats: &mut Stats, is_func_space: bool, is_unit: bool) -> (
     (start, end)
 }
 
+#[inline(always)]
+// This function discriminates among the comments that are next to code lines and
+// the ones that are on a single line. This difference is necessary in
```
marco-c, line 144: they are all on a "single" line, we should find a better way to describe them (here and in the name of the variable)

marco-c:
Can you test cases like:
```
   SOME CODE AND BLANK LINE
   printf("Ciao");  /* comment starts here
   continues here
   and ends here */
   MORE CODE
```


```
   SOME CODE AND BLANK LINE
   printf("Ciao");
   /* comment starts here
   continues here
   and ends here */
   MORE CODE
```

```
   SOME CODE AND BLANK LINE
   printf("Ciao");
   /* comment starts here
   continues here
   and ends here */ printf("Ciao;")
   MORE CODE
```

Luni-4:
> Can you test cases like:
> 
> ```
>    SOME CODE AND BLANK LINE
>    printf("Ciao");  /* comment starts here
>    continues here
>    and ends here */
>    MORE CODE
> ```
> 
> ```
>    SOME CODE AND BLANK LINE
>    printf("Ciao");
>    /* comment starts here
>    continues here
>    and ends here */
>    MORE CODE
> ```
> 
> ```
>    SOME CODE AND BLANK LINE
>    printf("Ciao");
>    /* comment starts here
>    continues here
>    and ends here */ printf("Ciao;")
>    MORE CODE
> ```

Done

src/metrics/loc.rs
```diff
@@ -142,6 +140,47 @@ fn init(node: &Node, stats: &mut Stats, is_func_space: bool, is_unit: bool) -> (
...
+    }
+}
+
+#[inline(always)]
+// Detects the comments that are on a code line but *before* the code part.
+// This difference is necessary in order to avoid having
+// a wrong count for the blank metric.
+fn cloc_before_code_line(stats: &mut Stats, start_code_line: usize) {
+    if let Some(end) = stats.comment_line_end {
+        if !stats.lines.contains(&start_code_line) && end == start_code_line {
```
marco-c, line 176: Move the `end == start_code_line` condition before `!stats.lines.contains(&start_code_line)`, since it is faster to evaluate

src/metrics/loc.rs
```diff
@@ -142,6 +140,47 @@ fn init(node: &Node, stats: &mut Stats, is_func_space: bool, is_unit: bool) -> (
...
 }
 
+#[inline(always)]
+// Discriminates among the comments that are *after* a code line and
+// the ones that are on an independent line.
+// This difference is necessary in order to avoid having
+// a wrong count for the blank metric.
+fn add_cloc_lines(stats: &mut Stats, start: usize, end: usize) {
+    let comment_diff = end - start;
+    let is_comment_afer_code_line = stats.lines.contains(&start);
```
marco-c, line 150:
```suggestion
    let is_comment_after_code_line = stats.lines.contains(&start);
```

marco-c:
Can you run the minimal tests script to see the difference before/after your patch on m-c? There might be cases we haven't thought of.

src/metrics/loc.rs
```diff
@@ -142,6 +140,47 @@ fn init(node: &Node, stats: &mut Stats, is_func_space: bool, is_unit: bool) -> (
...
+        // a comment *before* a code line is considered
+        stats.comment_line_end = Some(end);
+    }
+}
+
+#[inline(always)]
+// Detects the comments that are on a code line but *before* the code part.
+// This difference is necessary in order to avoid having
+// a wrong count for the blank metric.
+fn cloc_before_code_line(stats: &mut Stats, start_code_line: usize) {
```
marco-c, line 174: The name of this function is confusing, could you try to find a better name? Something like `check_comment_ends_on_code_line`

src/metrics/loc.rs
marco-c, line 150:
```suggestion
    let is_comment_after_code_line = stats.lines.contains(&start);
```

Luni-4:
@marco-c 

I can see only improvements and I don't find an additional useful test, have a loot at it also, perhaps I've lost something important https://community-tc.services.mozilla.com/api/queue/v1/task/MIBNQ1TgTlaK_x2yDtPw_g/runs/0/artifacts/public%2Fjson-diffs-and-minimal-tests.tar.gz

marco-c:
@Luni-4 looks good to me. Can you remove the test commit?

marco-c (approved):

PR #594, opened 2021-03-30
Delete Rust submodule

### LINKED ISSUE #442: Replace tree-sitter submodules with Rust bindings

Luni-4:
This PR removes the last submodule and fixes #442

### DIFF
--- a/enums/src/macros.rs
+++ b/enums/src/macros.rs
@@ -21,19 +21,12 @@ macro_rules! mk_get_language {
                   LANG::Tsx => tree_sitter_typescript::language_tsx(),
                   LANG::Javascript => tree_sitter_javascript::language(),
                   LANG::Python => tree_sitter_python::language(),
+                  LANG::Rust => tree_sitter_rust::language(),
                   LANG::Preproc => tree_sitter_preproc::language(),
                   LANG::Ccomment => tree_sitter_ccomment::language(),
                   LANG::Cpp => tree_sitter_mozcpp::language(),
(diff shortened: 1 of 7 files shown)


enums/src/macros.rs
```diff
@@ -15,24 +15,17 @@ macro_rules! mk_enum {
...
-                  LANG::Mozjs => tree_sitter_mozjs::language(),
-                  _ => match lang {
-                    $(
-                        LANG::$camel => {
-                            extern "C" { fn $name() -> Language; }
-                            unsafe { $name() }
-                        },
-                    )*
-                }
+            match lang {
```
marco-c, line 18: Could you keep the old indentation? Otherwise it looks like you're changing more than you actually are

marco-c (approved):

marco-c (approved):

PR #595, opened 2021-03-30
Pin tree-sitter lib and grammars

Luni-4:
This PR pins `tree-sitter` lib and grammars to avoid having differences in metrics without noticing them.


marco-c:
Looks good, but we need to do the same in the other Cargo.toml files we have around the repo.

I hope Dependabot will still notify us of updates.

marco-c (approved):

marco-c (approved):

PR #598, opened 2021-03-30
Improve generate grammars scripts

Luni-4:
This PR improves the scripts to generate the grammars in these ways:

- Retrieve the sha1 of the commits associated to the `tree-sitter-javascript` and `tree-sitter-mozcpp` crates from `crates.io` automatically, so it is no more necessary to update them manually.
- Use the script to recreate the grammars inside the script to generate the grammars

Commits should be reviewed one-by-one


marco-c (approved):
Way less problematic this way :+1:

PR #599, opened 2021-03-30
Do not notify submodules updates anymore on dependabot

Luni-4:
This PR avoids `dependabot` notifying a submodule update, since they are no longer supported


marco-c (approved):
This is not really necessary since Dependabot won't find any submodule to update anyway, but let's clean the config up anyway.

Issue #289, opened 2020-10-05
Use a stack approach for cognitive complexity


Luni-4:
Currently the cognitive complexity metric analyzes the nodes using ancestors. It would be helpful to analyze the nodes using a stack approach

Luni-4:
Change `else_if`s implementation too, as suggested in #603

Luni-4:
Closed in #935

PR #603, opened 2021-03-31
Implement an else-if checker for each grammar

### LINKED ISSUE #407: Code for is_else_if is probably wrong
For reference:
https://github.com/mozilla/rust-code-analysis/blob/fd55a85b8c0d70365e37961c743def4df3ccba4d/src/macros.rs#L20

There is no ref to `else` in the code so it's likely wrong.

Luni-4:
This PR implements a specific `else-if` checker for each grammar.

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -4,21 +4,35 @@ use regex::bytes::Regex;
 use crate::*;
 
 macro_rules! mk_else_if {
-    ($if_type: ident, $block_type: ident) => {
+    ($language: ident, $if_type: ident, $block_type: ident) => {
         #[inline(always)]
         fn is_else_if(node: &Node) -> bool {
-            if node.object().kind_id() != <Self as TSLanguage>::BaseLang::$if_type {
+            use $language::*;
+            if node.object().kind_id() != ElseClause {
                 return false;
             }
-            if let Some(parent) = node.object().parent() {
-                if parent.kind_id() == <Self as TSLanguage>::BaseLang::ElseClause {
+            // Index is 1 because the 0 one is the `else` keyword
+            if let Some(second_child) = node.object().child(1)
+            /* else, if_type/block_type */
+            {
+                if second_child.kind_id() == $if_type {
                     return true;
                 }
-                if let Some(grandparent) = parent.parent() {
-                    return grandparent.kind_id() == <Self as TSLanguage>::BaseLang::ElseClause
-                        && parent.kind_id() == <Self as TSLanguage>::BaseLang::$block_type
-                        && parent.child_count() == 3; // Grammars count { } as children too
+                if second_child.kind_id() != $block_type {
+                    return false;
                 }
+                let mut cursor = second_child.walk();
+                let mut is_else_if = true;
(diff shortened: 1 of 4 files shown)


src/checker.rs
```diff
@@ -291,5 +306,186 @@ impl Checker for RustCode {
...
+               time = 7;
+             }",
+            "foo.c",
+            CppParser,
+            true
+        );
+    }
+
+    #[test]
+    fn cpp_no_else_if() {
```
marco-c, line 392: Can you add this same test for all grammars?

src/checker.rs
```diff
@@ -117,6 +154,11 @@ impl Checker for JavaCode {
     mk_checker!(is_func, MethodDeclaration);
     mk_checker!(is_func_space, Program, ClassDeclaration);
     mk_checker!(is_non_arg,);
+
+    #[inline(always)]
+    fn is_else_if(_: &Node) -> bool {
+        false
+    }
```
marco-c, line 161: Let's use https://doc.rust-lang.org/std/macro.unimplemented.html in the unimplemented `is_else_if`s.

src/macros.rs
```diff
@@ -13,23 +13,6 @@ macro_rules! mk_checker {
     };
 }
 
-#[macro_use]
-macro_rules! mk_else_if {
```
marco-c, line 17: Can't we keep the macro here?

src/checker.rs
```diff
@@ -108,6 +140,11 @@ impl Checker for PythonCode {
     mk_checker!(is_func, FunctionDefinition);
     mk_checker!(is_func_space, Module, FunctionDefinition, ClassDefinition);
     mk_checker!(is_non_arg, LPAREN, COMMA, RPAREN);
+
+    #[inline(always)]
+    fn is_else_if(node: &Node) -> bool {
+        node.object().kind_id() == <Self as TSLanguage>::BaseLang::ElifClause
```
calixteman, line 146: I think `else if` is legal too in python

src/checker.rs
```diff
@@ -3,6 +3,28 @@ use regex::bytes::Regex;
...
+        #[inline(always)]
+        fn is_else_if(node: &Node) -> bool {
+            if node.object().kind_id() != <Self as TSLanguage>::BaseLang::$if_type {
+                return false;
+            }
+            if let Some(parent) = node.object().parent() {
+                if parent.kind_id() == <Self as TSLanguage>::BaseLang::ElseClause {
+                    return true;
+                }
+                if let Some(grandparent) = parent.parent() {
```
calixteman, line 17: I really wonder if it's good way to do...
To be clear I don't say that this code is bad but I've just a bad feeling (and maybe I'm totally wrong).
I need think about that.

Luni-4: Yes, the implementation should better consider how an AST is visited, but a different one would require to fix #289 first. Since we have added tests which seem to work correctly, I propose to postpone a more correct implementation at a later time.

Luni-4:
No more necessary because of #935

PR #609, opened 2021-04-08
Publish moz-grammars automatically on crates.io

### LINKED ISSUE #608: Pusblish moz-grammars on crates.io automatically
It would be helpful to publish `moz-grammars` on [crates.io](https://crates.io/) automatically using the CI

Luni-4:
This PR publishes `moz-grammars` on [crates.io](https://crates.io/) automatically when a new release is being created. It fixes #608

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -331,7 +331,7 @@ tasks:
               features:
                 taskclusterProxy: true
               maxRunTime: 3600
-              image: "mozilla/taskboot:0.3.2"
+              image: "mozilla/taskboot:0.3.3"
               env:
                 TASKCLUSTER_SECRET: project/relman/rust-code-analysis/deploy
               command:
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -341,6 +341,10 @@ tasks:
                    git config --global user.name moz.tools Bot &&
                    git clone --quiet ${repository} &&
                    cd rust-code-analysis &&
+                   cd tree-sitter-ccomment && taskboot cargo-publish && cd .. &&
+                   cd tree-sitter-preproc && taskboot cargo-publish && cd .. &&
+                   cd tree-sitter-mozjs && taskboot cargo-publish && cd .. &&
+                   cd tree-sitter-mozcpp && taskboot cargo-publish && cd .. &&
```
marco-c, line 347: What happens if the release already exists on crates.io?

Luni-4: `taskboot cargo-publish` does not discriminate among errors, so if a package is already published, it returns an error and the CI is interrupted

marco-c: OK, so this way we need to bump the versions of those crates even if nothing changed within them.

marco-c: WFM

marco-c (approved):

PR #611, opened 2021-04-08
Support m-c macros where first argument is a declaration

calixteman:
It aims to fix:
https://bugzilla.mozilla.org/show_bug.cgi?id=1702411#c8
I regenerated mozcpp grammar using ts-cpp and ts-c from their master versions and in using ts 0.19.4.
Tell me if I should change something about versions here.


janvarga:
I would also add `QM_WARNONLY_TRY_UNWRAP` to the list.

marco-c:
@calixteman LGTM, can you update the version of the package at https://github.com/mozilla/rust-code-analysis/blob/07784948bbbeceae158bb6adecfcc5222eca3da2/Cargo.toml#L40 and at https://github.com/mozilla/rust-code-analysis/blob/07784948bbbeceae158bb6adecfcc5222eca3da2/tree-sitter-mozcpp/Cargo.toml#L4?

This way CI will run RCA on all files in mozilla-central and will output a list of differences.

Luni-4 (changes requested):
Seems good to me!

Please update `tree-sitter-mozcpp` version in `enums/Cargo.toml` too

marco-c (approved):

PR #618, opened 2021-04-10
Update dependencies

calixteman:
@marco-c could you check that everything is fine with bugbug since actix is updated here ?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,7 +14,7 @@ dependencies = [
  "futures-sink",
  "log",
  "pin-project 0.4.28",
- "tokio",
+ "tokio 0.2.25",
  "tokio-util",
 ]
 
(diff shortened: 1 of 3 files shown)


marco-c:
@calixteman this looks similar to 72317cebd8097f0782dd820532d87dc2ff994811, but I had to revert it because the web server was not working anymore (it was panicking at start). Is it not the case anymore?

marco-c:
Yeah, I'm seeing:
```
thread 'main' panicked at 'System is not running', /home/marco/.cargo/registry/src/github.com-1ecc6299db9ec823/actix-rt-1.1.1/src/system.rs:78:21
```
when I run `cargo run -- --port 8080`.

marco-c:
It's probably some version incompatibility, because in the Cargo.lock diff I see some packages now depending on `actix-rt 2.2.0` and some on `actix-rt 1.1.1`.

calixteman:
Setting `actix-web = "4.0.0-beta.5"` fixes the problem.
We should wait for a new release, wdyt ?

marco-c:
>  We should wait for a new release, wdyt ?

WFM

PR #622, opened 2021-04-13
Allow grammar's AST nodes to be printed as strings 

### LINKED ISSUE #621: Swap keys and values in `phf` maps used to define the grammar of a language
To implement #232, it is necessary to swap the keys and values of the `phf` maps used to define the grammar of a language. Actually they are unused.

Luni-4:
This PR fixes #621 allowing grammar's AST nodes to be printed as strings


marco-c:
I think it would help if you submit the change for #232 too (either as a separate commit in this PR, or as a new PR based on this one), otherwise it'll be hard for me to understand how you want to use this.

src/languages/language_ccomment.rs
```diff
@@ -21,38 +21,41 @@ pub enum Ccomment {
...
-            Ccomment::PreprocContinuationLine,
-        ),
+        ("Nothing", "nothing"),
+        ("RawStringLiteral", "raw_string_literal"),
+        ("StringLiteralToken1", "string_literal_token1"),
+        ("DefineToken1", "define_token1"),
+        ("CharLiteralToken1", "char_literal_token1"),
+        ("CharLiteral", "char_literal"),
+        ("Define", "define"),
+        ("TranslationUnitRepeat1", "translation_unit_repeat1"),
```
marco-c, line 42: What is the point of this mapping? E.g. can't you print `TranslationUnitRepeat1` instead of `translation_unit_repeat1`?

Luni-4:
This PR is now contained in #277, since it is part of that and it cannot be treated separately. So we can close this one

PR #623, opened 2021-04-14
Add a script to split minimal tests into distinct directories depending on metrics differences

Luni-4:
This PR adds to CI a script that splits minimal tests into distinct directories depending on metrics differences. 

Advantages of this script:
- A more precise subdivision of metrics in directories
- Cross-platform

### DIFF
--- a/check-grammars-crates.sh
+++ b/check-grammars-crates.sh
@@ -92,29 +92,12 @@ if [ "$(ls -A $COMPARE)" ]; then
     # Maximum number of considered minimal tests for a metric
     MT_THRESHOLD=30
 
-    # Array containing the considered metrics
-    # TODO: Implement a command into rust-code-analysis-cli that returns all
-    # computed metrics https://github.com/mozilla/rust-code-analysis/issues/478
-    METRICS=("cognitive" "sloc" "ploc" "lloc" "cloc" "blank" "cyclomatic" "halstead" "nom" "nexits" "nargs")
-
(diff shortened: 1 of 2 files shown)


check-grammars-crates.sh
```diff
@@ -107,8 +107,11 @@ if [ "$(ls -A $COMPARE)" ]; then
     for METRIC in "${METRICS[@]}"
     do
 
-        PREFIX_METRIC="\.$METRIC"
-        FILES=`grep -r -i -l $PREFIX_METRIC $COMPARE | head -$MT_THRESHOLD`
+        # Find metrics as whole words
+        MATCH_PATTERN="\b\.$METRIC\b"
+        # Exclude these patterns for metrics
```
marco-c, line 112: Why should this pattern be excluded?

Luni-4: Because in this way we exclude files containing the name of a metric inside minimal tests code. Just an example:
The `nom` metric is different, so we have an `.html` file. Inside this file we have a function called `blank()`. When minimal tests are divided according to metrics, we have the `nom` html file inside the blank directory, but the `blank` metric wasn't modified at all

marco-c: Can you expand the comment in the file with what you just said?

Luni-4: Sure :)

marco-c (approved):

PR #627, opened 2021-04-21
Split exit in distinct files depending on the grammar

Luni-4:
This PR splits `exit` implementation in different files depending on the grammars. In this way it is much easier to add specific code for a determined grammar.


Luni-4:
Thinking a better alternative to add specific code for a determined grammar, perhaps in `getter` or `checker` files

PR #669, opened 2021-09-15
Fix warnings and cliippy lints

Luni-4:
This PR fixes some compiler warnings and `Clippy` lints


src/languages/mod.rs
```diff
@@ -1,3 +1,7 @@
+// FIXME: this should be fixed in some way to avoid useless allocations
+#![allow(clippy::cmp_owned)]
+#![allow(clippy::enum_variant_names)]
```
marco-c, line 3: Can you file an issue about these so we don't forget?

marco-c (approved):

PR #670, opened 2021-09-15
Fix mdbook building on CI

Luni-4:
This PR fixes the  mdbook building on CI. 

The `search` feature has been removed from mdbook, so building the tool with it leads to some dependency conflicts


marco-c (dismissed):

marco-c:
Could you write a better commit message?
"Fix documentation CI" doesn't explain what the problem is.

marco-c:
The commit message still doesn't explain what the problem was and how it was fixed :P

Luni-4:
> The commit message still doesn't explain what the problem was and how it was fixed :P

Ok, now it should be fixed :)

marco-c (approved):

PR #676, opened 2021-09-22
Move c_macros generation into the enums crate

Luni-4:
- Move  `c_macros` generation  to the `enums` crate in order to avoid recreating each `c_macro` file whenever the `rust-code-analysis` library is built.
- Replace `phf_set` with a const array


marco-c:
Can we remove the `phf` dependency as part of this or is it still necessary?

Luni-4:
> Can we remove the `phf` dependency as part of this or is it still necessary?

I would have done that in a following PR, but I can remove it in this PR

marco-c:
> I would have done that in a following PR, but I can remove it in this PR

Yes, thanks, I'd prefer that

marco-c:
This looks good to me, @calixteman do you want to take a look?

calixteman (approved):
LGTM too. r+
Thanks for doing that.

marco-c (approved):

Issue #359, opened 2020-11-16
Implement metrics for Java code


Luni-4:
Currently any metrics have been implemented for the `Java` code

marco-c:
As part of fixing this, we can write some docs on how to add a new language.

dburriss:
I would like to pick this up but not too sure what it entails. I have been looking through the implementations and think I understand some of it. I have only been learning Rust for a few weeks so it might take a while but since this has been open for almost a year I imagine it is not urgent. Eventually, I would like to add Go and C# but it appears Java has some implementation already so I thought this would be a good place to start.

I could start with investigating the different elements that need implementing and documenting those if it does not exist already? This could for the basis of documentation mentioned.

dburriss:
Entirely possible I am missing something here and this all exists but the metrics seem to rely on some sort of language grammar being generated?
This is done by adding the necessary tree_sitter lib to `rust-code-analysis/tree/master/enums/src` and adding it to the necessary language call and macro? This seems implemented for Java so not a big deal but I would like to circle around eventually and tackle Go (seems to be some mentions in the codebase) and C# (nothing).

I imagined it is going to be fairly iterative.

A few docs I can imagine:
- High level how-to on language going over the different moving parts that need implementing
- Metrics Reference (point to existing docs)
- Tutorial style implementation (here is one I did for Farmer https://compositionalit.github.io/farmer/contributing/adding-resources/ )

With the rough sketch of how things work done here, I write it up so I am explaining it back to you to be sure I understand. Then I implement and come back and flesh out the docs with more detail as I learn.

Of course with enough help I am sure we could document it all up front, it just was not how I imagined doing it when I started poking around.

Something I probably need to look into is more on how Rust generates its docs.

Is that far off what you imagined?  
P.S. I am going to be AFK for 2 weeks so will pick this up when I am back online.

dburriss:
Hi. Working on this again.
I am having some trouble with something but want to check it is something I should be solving.

```java
if (x == 1)
```
Is one logical line.

```java
if ( (x == 1) || (x==2) )
```
I had assumed this was 2 logical lines but reading [the link](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/expressions.html) from @Luni-4 I am thinking I may have been wrong. Although it never talks about the semantics of a compound predicate, my understanding now is that it is a single expression and so a single logical line. Do you agree?

marco-c:
[...]

PR #694, opened 2021-10-26
Java LoC metric

dburriss:
This PR is for adding the LoC metric discussed in issue #359.

### DIFF
--- a/src/getter.rs
+++ b/src/getter.rs
@@ -515,6 +515,9 @@ impl Getter for JavaCode {
 
         let typ = node.object().kind_id();
         match typ.into() {
+            ClassDeclaration => SpaceKind::Class,
+            MethodDeclaration | LambdaExpression => SpaceKind::Function,
+            PackageDeclaration | ModuleDeclaration => SpaceKind::Namespace,
             Program => SpaceKind::Unit,
             _ => SpaceKind::Unknown,
         }
--- a/src/getter.rs
+++ b/src/getter.rs
@@ -515,9 +515,8 @@ impl Getter for JavaCode {
 
         let typ = node.object().kind_id();
         match typ.into() {
-            ClassDeclaration => SpaceKind::Class,
+            InterfaceDeclaration | ClassDeclaration => SpaceKind::Class,
             MethodDeclaration | LambdaExpression => SpaceKind::Function,
-            PackageDeclaration | ModuleDeclaration => SpaceKind::Namespace,
             Program => SpaceKind::Unit,
             _ => SpaceKind::Unknown,
         }
--- a/src/getter.rs
+++ b/src/getter.rs
@@ -509,4 +509,17 @@ impl Getter for CppCode {
 
 impl Getter for PreprocCode {}
 impl Getter for CcommentCode {}
-impl Getter for JavaCode {}
+impl Getter for JavaCode {
+    fn get_space_kind(node: &Node) -> SpaceKind {
+        use Java::*;
+
+        let typ = node.object().kind_id();
+        match typ.into() {
+            MethodDeclaration | LambdaExpression => SpaceKind::Function,
+            Class | ClassDeclaration => SpaceKind::Class,
+            PackageDeclaration => SpaceKind::Namespace,
+            Program => SpaceKind::Unit,
+            _ => SpaceKind::Unknown,
+        }
+    }
+}
--- a/src/getter.rs
+++ b/src/getter.rs
@@ -515,9 +515,6 @@ impl Getter for JavaCode {
 
         let typ = node.object().kind_id();
         match typ.into() {
-            MethodDeclaration | LambdaExpression => SpaceKind::Function,
-            Class | ClassDeclaration => SpaceKind::Class,
-            PackageDeclaration => SpaceKind::Namespace,
             Program => SpaceKind::Unit,
             _ => SpaceKind::Unknown,
         }
--- a/src/metrics/loc.rs
+++ b/src/metrics/loc.rs
@@ -810,12 +810,19 @@ impl Loc for CppCode {
     }
 }
 
+// impl fmt::Display for Java {
+//     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+//        write!(f, "{:?}", self)
+//     }
+// }
+
 impl Loc for JavaCode {
     fn compute(node: &Node, stats: &mut Stats, is_func_space: bool, is_unit: bool) {
         use Java::*;
 
         let (start, end) = init(node, stats, is_func_space, is_unit);
-        let kind_id = node.object().kind_id().into();
+        let kind_id : Java = node.object().kind_id().into();
+        //println!("KINDID {}", kind_id.to_string());
         match kind_id {
             Program => {}
             Comment => {
@@ -823,11 +830,11 @@ impl Loc for JavaCode {
             }
             AssertStatement
             | BreakStatement
-            | MethodInvocation
+            | BinaryExpression
             | ContinueStatement
             | DoStatement
+            | Declaration
             | ExpressionStatement
-            | ForStatement
             | IfStatement
             | LocalVariableDeclaration
             | ReturnStatement
@@ -836,8 +843,20 @@ impl Loc for JavaCode {
             | TernaryExpression
             | ThrowStatement
             | TryStatement
+            | UpdateExpression
             | WhileStatement => {
                 stats.logical_lines += 1;
+            }
+            For => {
+                if count_specific_ancestors!(
+                    node,
+                    ForStatement, Block 
+                ) == 0
+                {
+                    // handle for(int i:arr)
+                    // otherwise the statements in the for are counted elsewhere
+                    stats.logical_lines += 1;
+                }
             }
             _ => {
                 check_comment_ends_on_code_line(stats, start);
@@ -2058,10 +2077,10 @@ mod tests {
     }
 
     #[test]
-    fn java_simple_lloc() {
+    fn java_for_lloc() {
         check_metrics!(
-            "for (int i = 0; i < 100; i++) {
-               System.out.println(i);
+            "for (int i = 0; i < 100; i++) { // + 3
+               System.out.println(i); // + 1
              }",
             "foo.java",
             JavaParser,
@@ -2073,12 +2092,29 @@ mod tests {
     }
 
     #[test]
+    fn java_foreach_lloc() {
+        check_metrics!(
+            "
+            int arr[]={12,13,14,44}; // +1
+            for (int i:arr) { // +1
+               System.out.println(i); // +1
+             }",
+            "foo.java",
+            JavaParser,
+            loc,
+            [
+                (lloc, 3, usize), // The number of statements is 3
+            ]
+        );
+    }
+
+    #[test]
     fn java_multi_lloc() {
         check_metrics!(
-            "int max = 10;
+            "int max = 10; // +1
 
-            for (int i = 0; i < max; i++) {
-               System.out.println(i);
+            for (int i = 0; i < max; i++) { // +3
+               System.out.println(i); // +1
              }",
             "foo.java",
             JavaParser,
@@ -2164,7 +2200,7 @@ mod tests {
             
             class HelloWorldApp {
               public void main(String[] args) {
-                System.out.println(\"Hello World!\"); // Display the string.
+                System.out.println(\"Hello World!\"); // Display the string. Only statement. +1 lloc
               }
             }",
             "foo.java",
@@ -2173,7 +2209,7 @@ mod tests {
             [
                 (sloc, 11, usize), // The number of lines is 11
                 (ploc, 6, usize),  // The number of code lines is 5
-                (lloc, 2, usize),  // The number of statements is 2
+                (lloc, 1, usize),  // The number of statements is 1
                 (cloc, 5, usize),  // The number of comments is 5
                 (blank, 1, usize)  // The number of blank lines is 1
             ]
--- a/src/metrics/loc.rs
+++ b/src/metrics/loc.rs
@@ -822,6 +822,7 @@ impl Loc for JavaCode {
 
         let (start, end) = init(node, stats, is_func_space, is_unit);
         let kind_id: Java = node.object().kind_id().into();
+        println!("{}", kind_id);
         match kind_id {
             Program => {}
             Comment => {
@@ -831,26 +832,27 @@ impl Loc for JavaCode {
             | AssignmentExpression
             | BreakStatement
             | BinaryExpression
+            | CastExpression
             | ContinueStatement
             | DoStatement
             | Declaration
+            | EnhancedForStatement
             | IfStatement
+            | InstanceofExpression
+            | LambdaExpression
             | LocalVariableDeclaration
             | MethodInvocation
+            | ParenthesizedExpression
+            | PrimaryExpression
             | ReturnStatement
             | Statement
             | SwitchStatement
+            | TernaryExpression
             | ThrowStatement
             | TryStatement
-            | UpdateExpression => {
+            | UpdateExpression
+            | UnaryExpression => {
                 stats.logical_lines += 1;
-            }
-            For => {
-                if count_specific_ancestors!(node, ForStatement, Block) == 0 {
-                    // handle for(int i:arr)
-                    // otherwise the statements in the for are counted elsewhere
-                    stats.logical_lines += 1;
-                }
             }
             _ => {
                 check_comment_ends_on_code_line(stats, start);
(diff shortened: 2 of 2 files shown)


marco-c:
@dburriss still interested in finishing this up?

dburriss:
Yes. Just not a lot of time at the moment. Hopefully be able to come back to this in a week or 2.

dburriss:
@marco-c @Luni-4 this is up-to-date with master and ready for review.

src/getter.rs
```diff
@@ -509,4 +509,17 @@ impl Getter for CppCode {
...
 impl Getter for CcommentCode {}
-impl Getter for JavaCode {}
+impl Getter for JavaCode {
+    fn get_space_kind(node: &Node) -> SpaceKind {
+        use Java::*;
+
+        let typ = node.object().kind_id();
+        match typ.into() {
+            MethodDeclaration | LambdaExpression => SpaceKind::Function,
+            Class | ClassDeclaration => SpaceKind::Class,
```
calixteman, line 519: Did you intentionally miss interface:
https://github.com/tree-sitter/tree-sitter-java/blob/a24ae7d16de3517bff243a87d087d0b4877a65c5/grammar.js#L632
?
If yes, please add a comment to explain why.
In looking at the above link I can see that there is a `module` concept (I don't know what is: I didn't write Java code for years), so same question.

src/metrics/loc.rs
```diff
@@ -500,9 +500,59 @@ impl Loc for CppCode {
     }
 }
 
+impl fmt::Display for Java {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?}", self)
```
calixteman, line 505: I don't see similar code for other languages, is it a debug leftover ?
If it's useful in general, I guess it can be added for all languages in a follow-up PR.

src/metrics/loc.rs
```diff
@@ -500,9 +500,59 @@ impl Loc for CppCode {
...
+        use Java::*;
+
+        let (start, end) = init(node, stats, is_func_space, is_unit);
+        let kind_id: Java = node.object().kind_id().into();
+        match kind_id {
+            Program => {}
+            Comment => {
+                add_cloc_lines(stats, start, end);
+            }
+            AssertStatement
```
calixteman, line 520: Maybe you missed `InstanceofExpression` and few others, have a look on:
https://github.com/tree-sitter/tree-sitter-java/blob/a24ae7d16de3517bff243a87d087d0b4877a65c5/grammar.js#L161-L170

src/getter.rs
dburriss: Good catch. Since this is my first time working in Rust, this repo, and tree-sitter, rather than copying other implementations I just added grammar to make tests pass. 
I will take a scan through to see if any other things I missed.
It has been a decade or so since I have looked at Java... I do not remember ever using `module` :D I learned something today.

src/metrics/loc.rs
dburriss: It indeed is from debugging. Ok to leave in then? Should I create an issue to track adding it to the others? Guss can discuss there and if we decide no, then I will remove this.

calixteman: I would prefer having that in an other PR for all the different languages.

dburriss: Will do. I will pull it out of this PR once the other items in this are sorted out. I have not found a better way of seeing what is going on.

src/getter.rs
dburriss: For the SLOC off by 1 issue the only match I need is `Program => SpaceKind::Unit`. The space_kind for the rest can be handled in the implementations that need them. Otherwise no test coverage. Does that make sense to you @calixteman ?

calixteman: What's the SLOC off by 1 issue ?

src/metrics/loc.rs
calixteman: @dburriss ^^

src/metrics/loc.rs
```diff
@@ -1428,4 +1477,284 @@ mod tests {
...
+        check_metrics!(
+            "package com.company;
+             /**
+             * The HelloWorldApp class implements an application that
+             * simply prints \"Hello World!\" to standard output.
+             */
+            
+            class HelloWorldApp {
+              public void main(String[] args) {
+                String message = args.length == 0 ? \"Hello empty world\" : \"Hello world\"; // +2 lloc : 1 var assignment + binary exp
```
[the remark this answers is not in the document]
calixteman: yes it does.

src/metrics/loc.rs
dburriss: @calixteman  I should have mentioned it explicitly. I did add these in but it completely breaks the counts. I will record some examples. Maybe we can get to the same counts with fewer statements and more expressions.

src/getter.rs
dburriss: Well, I left this note for myself while documenting my changes. And no I don't generally make my notes as if I am talking to myself but I had initially thought to add to the documentation as I went.

> As you implement tests, you may find yourself running into an off by 1 error for SLOC count. 
> For most code we expect the `SpaceKind` to be `SpaceKind::Unit` but the default is `SpaceKind::Unknown`.
> To solve this you need to implement the `Getter` trait in [src/getter.rs](https://github.com/mozilla/rust-code-analysis/blob/master/src/getter.rs). Specifically, the `get_space_kind` function.

I think it is because any java snippet of code gets a `Program` leaf. I guess `UNKNOWN` gets counted and `UNIT` does not? Or maybe the other way around? It was a few months ago that I wrote it...

src/metrics/loc.rs
dburriss: For ease of discussion, here is the result of parsing that:
```
LocalVariableDeclaration
TypeIdentifier
VariableDeclarator
Identifier
EQ
TernaryExpression
BinaryExpression
ParenthesizedExpression
LPAREN
BinaryExpression
Identifier
PERCENT
DecimalIntegerLiteral
RPAREN
EQEQ
DecimalIntegerLiteral
QMARK
StringLiteral
COLON
StringLiteral
SEMI
Comment
```

calixteman: I re-read the sloc & lloc definitions on wikipedia:
https://en.wikipedia.org/wiki/Source_lines_of_code#Measurement_methods
and in lloc we should have only statements.
I wonder why I talked about InstanceofExpression... maybe because there was an Expression here in the 1st version of your patch.... or maybe I was a bit tired... Anyway sorry for that.
For example with js we only have statements for logical_lines:
https://github.com/mozilla/rust-code-analysis/blob/d488b6eb5b4de522700a75a84602d404a5f64de1/src/metrics/loc.rs#L343-L349
But for rust code:
https://github.com/mozilla/rust-code-analysis/blob/d488b6eb5b4de522700a75a84602d404a5f64de1/src/metrics/loc.rs#L424-L436
the term `Expression` is used in place of `Statement`: it's a bit confusing.
So it really depends on the way the grammar is built itself.

Anyway, for Java we've a a grammar definition for statement:
https://github.com/tree-sitter/tree-sitter-java/blob/a24ae7d16de3517bff243a87d087d0b4877a65c5/grammar.js#L415
so I guess we can basically re-use what we've here and nothing else (like for js, python stuff).

src/metrics/loc.rs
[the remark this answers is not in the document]
calixteman: So after reading lloc doc on wikipedia: we should have 1 here: for all the assignment, which should be an ExpressionStatement.

src/metrics/loc.rs
dburriss: Yeah definitely depends on the language. I guess Rust takes some FP cues so expressions are preferred. Statements were the original push as discussed with Luni-4 https://github.com/mozilla/rust-code-analysis/issues/359#issuecomment-952631456. Sorrry, I should have thought to link to it sooner.
I'll revert back in the next day or 2. I learned a bit about Java. Have not touched it since version 6 so if nothing else I learned a few things about how it does lambdas. I must admit, I like it even less than I like its generics but hey. Each to their own.

src/getter.rs
dburriss: @calixteman does this one make sense? I imagine the rest will be added for other metrics as needed. Possibly needed for other metrics counting functions? I have not looked at anything but LoC and the Java Cyclomatic PR.

src/metrics/loc.rs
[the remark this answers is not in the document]
dburriss: @calixteman I missed this comment but I had a similar thought. I went with `ExpressionStatement` instead of method invocation and a few other things like handling `i++;`.
This specific line doesn't parse as an expression statement but does count as 1 LLOC as it is a `LocalVariableDeclaration`.

dburriss:
@calixteman A quick summary of the latest changes in commit "Remove expressions from count".

- I removed ALL Expression matches
- IMPORTANT: with the removal of `BinaryExpression` `UpdateExpression` means the numbers for a `for` would drop for 3 to 2. I will call this out with a comment specifically.
- updated the expectations in the tests from the expression changes. The major change in count is around ternary as is expected. 
- added some tests and changed a few others to cover `try`, `switch`, `do..while`, `throw`, `break`, and `continue`.

A reminder: I will remove the debug stuff once we are happy with the implementation.

src/metrics/loc.rs
dburriss: All expressions are removed now.

src/metrics/loc.rs
```diff
@@ -500,9 +500,55 @@ impl Loc for CppCode {
...
+            | SwitchStatement
+            | ThrowStatement
+            | TryStatement
+            | WhileStatement => {
+                stats.logical_lines += 1;
+                println!("+1");
+            }
+            ForStatement => {
+                // handle binary check + increment
+                stats.logical_lines += 2;
```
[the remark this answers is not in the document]
calixteman: mmh... on wikipedia:
https://en.wikipedia.org/wiki/Source_lines_of_code
we have 1 for the `for` statement.
Tbh, from our pov, we don't care that much: we use these numbers in some machine learning stuff so I guess it doesn't make any difference at the end.
So I think we must be consistent in the way to compute those numbers across the different languages.
@marco-c, @Luni-4 do we have, in the scientific literature, some clear definitions ?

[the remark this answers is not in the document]
dburriss: @Luni-4 Thanks. A good suggestion to link in the code. I will add that.
I did read the article when you linked to it in [this discussion](https://github.com/mozilla/rust-code-analysis/issues/359#issuecomment-952631456) and we agreed then that the classic `for` loop had a LLOC of 3, in the same comment.

From that Oracle article:

> Assignment expressions
> Any use of ++ or --
> Method invocations
> Object creation expressions

Not counting the binary expression but counting the `for` loop itself (like is done with `while` and `do...while`), still puts us at a count of 3 if we include the `++` as the article suggests. This all depends if we decide that what is inside a `for` loop is subject to those rules.

[This article from Oracle](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/for.html) refers to the initializer, condition, and incrementer as *expressions*. So since it is specific to the case we are talking about, my suggestion is we do as @Luni-4 suggests and make any `for` count as ` LLOC. That simplifies the code AND seems to be the most straightforward way to interpret those articles taken together.

So. `for` = 1 LLOC. Agreed?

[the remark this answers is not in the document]
Luni-4: Since Java considers a `for` as a single statement and its internals as expressions, I agree with the `for` = 1 LLOC line. Let's link both the articles in the code as comments though, so we know we have decided one way instead of the other

dburriss:
Ok in this latest push I have incorporated the changes from the comments from @Luni-4 (hopefully @calixteman you agree with the arguments)

- `for` statement is ` LLOC and so expressions inside are ignored
- commented with links to the Oracle articles on why we only count statements, including the one about `for` statements
- removed the debug lines etc.
- rebased the cyclomatic complexity and other changes from the past week or 2

Hopefully, it is ready to go.

dburriss:
@calixteman would you have a chance to review this before the weekend? That way I can make any requested changes over the weekend or move on to another.

marco-c:
@dburriss Calixte is off this week, perhaps you could start looking at C#?

marco-c:
Calixte was off this week too unfortunately, he's back today :)
BTW, you could join our Matrix room on https://chat.mozilla.org/#/room/#rust-code-analysis:mozilla.org.

dburriss:
Rebased in recent merges and fixed errors from recent `Stats` changes.

src/metrics/loc.rs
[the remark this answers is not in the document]
calixteman: yep, agreed too.

src/getter.rs
```diff
@@ -509,4 +509,14 @@ impl Getter for CppCode {
...
 impl Getter for PreprocCode {}
 impl Getter for CcommentCode {}
-impl Getter for JavaCode {}
+impl Getter for JavaCode {
+    fn get_space_kind(node: &Node) -> SpaceKind {
+        use Java::*;
+
+        let typ = node.object().kind_id();
+        match typ.into() {
+            Program => SpaceKind::Unit,
```
calixteman, line 518: My definition of space is: something which can have its own metrics.
So for example, you can have a class A with an inner class B each of them could be a space: this way we'll have some metrics for both of them:
 - B has its own metrics
 - A has its own metrics merged with the B ones
And at the end in the final output you'll have the granularity the user wants.
So not having more spaces here means that we don't have any granularity.

calixteman:
Sorry for the long delay: I've been to be off in the last weeks (life....).
Overall it looks good, just need to address the space thing issue and then I guess it'll be mergeable.

src/getter.rs
dburriss: Interesting. I have taken a shot at it based on the other implementations and that explanation. Other than that not having `Program => SpaceKind::Unit` causes the line count to be off (javascript and typescript have this match too), I am uncertain how to actually test this in the context of LoC?

dburriss: @calixteman I think this covers what can contain other code in Java. As I said in the above comment, I am uncertain how to test this in the context of LoC.

```rust
ClassDeclaration => SpaceKind::Class,
MethodDeclaration | LambdaExpression => SpaceKind::Function,
PackageDeclaration | ModuleDeclaration => SpaceKind::Namespace,
Program => SpaceKind::Unit
```

calixteman: If you want to see space in actions you could just run `cargo run -p rust-code-analysis-cli -- -m -p foo.java` and have a look on the output, with foo.java:
```java
class Outer {
  String x;
 
  class Inner {
    String y;
    int bar() {
       return 12;
    }
  }
}
```

If you return unknown for any kind of space, you'll get some `unknown` in the output.
Maybe the "off by 1" issue is related to:
https://github.com/mozilla/rust-code-analysis/blob/master/src/metrics/loc.rs#L40

dburriss: Ah cool! Makes complete sense. And yes that line of code was what led me to the fix of `Unit` for `Program` for the "off by 1" issue. It was in November last year when I started on this (and learning Rust) so the memories are a little hazy.

Adding a lambda expression and a package to the java file didn't seem to add anything to the structure. Not sure if I should leave them in, conceptually it seems correct to me as is. The code on this comment is not showing as outdated but I do think the kind is working as you wanted. See image:

image

src/getter.rs
```diff
@@ -509,4 +509,17 @@ impl Getter for CppCode {
...
 impl Getter for PreprocCode {}
 impl Getter for CcommentCode {}
-impl Getter for JavaCode {}
+impl Getter for JavaCode {
+    fn get_space_kind(node: &Node) -> SpaceKind {
+        use Java::*;
+
+        let typ = node.object().kind_id();
+        match typ.into() {
+            ClassDeclaration => SpaceKind::Class,
```
calixteman, line 518: Maybe you could add InterfaceDeclaration here too.
Maybe it's worthing adding a SpaceKind::Interface because an interface contributes to loc for example but it expected to have other metrics equal to 0 since it just some functions declarations.
So having interface in the output will help the user the understand this numbers and maybe (probably ?) skip this data since they aren't really useful (but maybe they're).

dburriss: I mapped `InterfaceDeclaration` as a `SpaceKind::Class` and it didn't show up at all in the hierarchy.

With AND without gave:

unit: foo.java
- function: print
- class: A6
-- function: print
-- function: main 

for the following code file:

```java
package com.java;

interface printable{  
  void print();  
}  

class A6 implements printable{  
  public void print(){System.out.println("Hello");}  
    
  public static void main(String args[]){  
    A6 obj = new A6();  
    obj.print();  
  }  
} 
```

Adding a `SpaceKind::Interface` interface could be interesting if it indeed showed up in the hierarchy. I would expect:

unit: foo.java
- interface:  printable
-- function: print
- class: A6
-- function: print
-- function: main 

Although as you said, not sure how many metrics interfaces actually contribute. Evaluating with "Lay of least surprise" I do think the `SpaceKind::Interface` makes sense but that should be a separate PR I would think?

calixteman (approved):
It looks good to me.
Thanks a lot for doing that.

Issue #410, opened 2021-01-06
Calculate min and max for all metrics


marco-c:
We are calculating sum and average, in some cases min and max could be useful too.

- [x] Cyclomatic
- [x] Nom
- [x] Loc
- [x] Cognitive
- [x] Nexit
- [x] Nargs
- [ ] Halstead
- [ ] Mi

Luni-4:
`Halstead` and `Mi` metrics depend on other metrics it is not so fundamental computing min and max operators for them

PR #699, opened 2021-11-18
Add min and max implementation for Cyclomatic

giovannitangredi:
This PR adds minimum and maximum value for the Cyclomatic metric, adding some new unit testing to handle some corner cases and partly resolving Issue #410 .


src/metrics/cyclomatic.rs
```diff
@@ -1,22 +1,27 @@
...
 use serde::Serialize;
 use std::fmt;
 
-use crate::checker::Checker;
-use crate::*;
-
 /// The `Cyclomatic` metric.
 #[derive(Debug, Clone)]
 pub struct Stats {
+    self_cyclomatic: f64,
```
calixteman, line 10: The role of this new field is unclear for me and its name isn't really self-explanatory.
So it'd be nice to add a comment to explain why we need it and why we can't simply use `cyclomatic`.

calixteman:
It would be nice (if it's possible) to add some references (papers, websites, ...) which could help to explain why min/max of cyclomatic is interesting. If they don't exist or if they aren't interesting, it isn't a problem, it's more out of curiosity (and I make some archeology on github I always appreciate to have some documentation (even if I don't do it myself all the time ;))).
Anyway we compute mean, so why not min/max, median, ... whatever I don't know.

src/metrics/cyclomatic.rs
calixteman: Sorry but the comment is still unclear for me:
 - the concept of `line 374` is a bit fuzzy in the time line, don't you think ?
 - my impress is that this field is useless (this is reinforced by the fact you're almost duplicating code for the both fields), and your comment doesn't explain anything I don't know, here you're telling me to read the c_unit_after test and try to figure out myself what's wrong when we don't save this self_cyclomatic... And if someone read your comment in 10 months he will have to spend some time to figure why this value is useful.
So here, you must help me to understand very quickly why you MUST save this value, with a small example or a small explanation.

src/metrics/cyclomatic.rs
```diff
@@ -80,10 +109,12 @@ impl Cyclomatic for PythonCode {
         match node.object().kind_id().into() {
             If | Elif | For | While | Except | With | Assert | And | Or => {
                 stats.cyclomatic += 1.;
+                stats.self_cyclomatic += 1.;
```
calixteman, line 112: You compute the same thing two times: `stats.cyclomatic += 1.;` and `stats.self_cyclomatic += 1.;`: maybe having a `self.self_cyclomatic = self.cyclomatic` at the right place would simplify things.

giovannitangredi: I will check if that is possible, but last time I checked I didn't find a way to avoid this change.

giovannitangredi:
I have done the following changes:
	-Removed self_cyclomatic.

	-Added a new field called cyclomatic_sum that stores the sum of all cyclomatic for each space merged, similar to the old usage of cyclomatic. Now this value is used as sum and it is also used for calculate average. This field is merged when the merge funciton is called.

	-Now cyclomatic only stores the cyclomatic value of the single space/method, it doesn't get modified by the merge. This value is used when computing min and max and it is added to the sum when the parsing of the space is completed (implemented in the compute_minmax function of cyclomatic). 

This changes  have been done because when we merge 2 states, the first one could still be uncompleted  and  so the min and max cannot be computed  correctly, now the min and max are only computed  when a state has been completed.

calixteman (approved):
LGTM. Thank you for doing that.

PR #709, opened 2021-11-26
Add APIs to process the files in a directory concurrently

Luni-4:
They can be used in different contexts, for example, to count the number of files in a directory, the number of its subdirectories or extract some information contained or scattered through different files.

They can also be used to compute some metrics that act on files in some way and integration tests that need repository to test things.

### DIFF
--- a/src/concurrent_files.rs
+++ b/src/concurrent_files.rs
@@ -1,0 +1,279 @@
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+use std::thread;
+
+use crossbeam::channel::{unbounded, Receiver, Sender};
+use globset::GlobSet;
+use walkdir::{DirEntry, WalkDir};
+
+type ProcFilesFunction<Config> = dyn Fn(PathBuf, &Config) -> std::io::Result<()> + Send + Sync;
+
+type ProcDirPathsFunction<Config> =
+    dyn Fn(&mut HashMap<String, Vec<PathBuf>>, &Path, &Config) + Send + Sync;
+
+type ProcPathFunction<Config> = dyn Fn(&Path, &Config) + Send + Sync;
+
+// Null functions removed at compile time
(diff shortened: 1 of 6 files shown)


src/concurrent_files.rs
```diff
@@ -0,0 +1,281 @@
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+use std::thread;
+
+use crossbeam::channel::{unbounded, Receiver, Sender};
+use globset::GlobSet;
+use walkdir::{DirEntry, WalkDir};
+
+type ProcFiles<Config> = Box<dyn Fn(PathBuf, &Config) -> std::io::Result<()> + Send + Sync>;
```
calixteman, line 10: This is only used in `ConcurrentRunner` definition and you use the same name (i.e. `ProcFiles`) here:
https://github.com/mozilla/rust-code-analysis/pull/709/files#diff-8c879faf238b8da3dbf21276b270b4832682947ff73ad1a1607a1371f89f66daR29
which is a bit confusing.
Wouldn't it be simpler to use:
`type ProcFiles = Fn(PathBuf, &Config) -> std::io::Result<()> + Send + Sync;`
with a better name to remember that's a function type ?

src/concurrent_files.rs
```diff
@@ -0,0 +1,281 @@
...
+        }
+    }
+}
+
+fn send_file<T>(
+    path: PathBuf,
+    cfg: &Arc<T>,
+    sender: &JobSender<T>,
+) -> Result<(), ConcurrentErrors> {
+    if let Err(e) = sender.send(Some(JobItem {
```
calixteman, line 52: nit: use `or_else`

calixteman (approved):
LGTM.
Thank you.

calixteman:
Could you fix the clippy issue before merging please ?

Luni-4:
> Could you fix the clippy issue before merging please ?

Done

calixteman (approved):

PR #710, opened 2021-11-26
Remove enum_iterator

lu-zero:
It is used only in mk_enum and the macro knows already its variants so it can be manually implemented.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -693,26 +693,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "enum-iterator"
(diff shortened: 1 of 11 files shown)


Luni-4 (changes requested):
Thanks a lot!

Please add the updated `enums/Cargo.lock` too and there should be everything!

calixteman (approved):
It removes a dependency so r+ for me, thx for doing this.

calixteman:
oh I missed Luni's comment (mid-air collision).
Please make another PR asap.

PR #712, opened 2021-11-26
Add  minimum and maximum implementation for the Nexit metric

giovannitangredi:
This PR adds minimum and maximum value for the Nexit metric, the method used is similar to min and max for cyclomatic metric, Tests were modified accordingly. 
Partly resolving Issue #410 .


src/metrics/cyclomatic.rs
```diff
@@ -32,7 +32,7 @@ impl Serialize for Stats {
     where
         S: Serializer,
     {
-        let mut st = serializer.serialize_struct("cyclomatic", 2)?;
+        let mut st = serializer.serialize_struct("cyclomatic", 4)?;
```
calixteman, line 35: Good catch

calixteman (approved):
With a green CI, r=me.

Issue #389, opened 2020-12-03
Add more integration tests for the lib


Luni-4:
We should add more integration tests for the languages we are already providing

Luni-4:
We could create a `data` directory containing a subdirectory for each programming language. In each of this subdirectories, it is present a repository of a medium-size project. For example:
- `data/javascript` ---> [pdf.js](https://github.com/mozilla/pdf.js)
- `data/cpp --> [deepspeech](https://github.com/mozilla/DeepSpeech)
- `data/rust` ---> [grcov](https://github.com/mozilla/grcov)

and so on

Each repository is pinned to a specific version and they are treated as git submodules.

The `tests` directory contains an integration test for each subdirectory and it can verify different properties:
- See whether the metrics are computed and their comparison is correct
- See if a project does not contain error nodes

Luni-4:
PR #724 added integration tests for the following languages:
- Rust
- Javascript
- C/C++

Other integration tests need to be added:
- [ ] Java
- [ ] TypeScript

PR #724, opened 2021-12-15
Add more integration tests

marco-ballario:
This PR adds DeepSpeech and grcov as git submodules, their json metrics files and 2 integration tests.
The tests compare rca output for the submodules with their json files previously generated.
To do that I used the APIs to process files in a directory concurrently introduced here: https://github.com/mozilla/rust-code-analysis/pull/709
It partly resolves issue #389

### DIFF
--- a/src/tools.rs
+++ b/src/tools.rs
@@ -229,13 +229,17 @@ pub fn guess_language<P: AsRef<Path>>(buf: &[u8], path: P) -> (Option<LANG>, Str
     }
 }
 
+/// Replaces \n and \r ending characters with a single generic \n
 pub(crate) fn remove_blank_lines(data: &mut Vec<u8>) {
-    let count_trailing = data.iter().rev().take_while(|&c| *c == b'\n').count();
+    let count_trailing = data
+        .iter()
+        .rev()
+        .take_while(|&c| (*c == b'\n' || *c == b'\r'))
+        .count();
     if count_trailing > 0 {
-        data.truncate(data.len() - count_trailing + 1);
-    } else {
-        data.push(b'\n');
+        data.truncate(data.len() - count_trailing);
     }
+    data.push(b'\n');
 }
 
 pub(crate) fn normalize_path<P: AsRef<Path>>(path: P) -> PathBuf {
--- a/tests/test.rs
+++ b/tests/test.rs
@@ -1,0 +1,276 @@
+use globset::GlobSet;
+use globset::{Glob, GlobSetBuilder};
+use rust_code_analysis::LANG;
+use rust_code_analysis::*;
+use std::fs::*;
+use std::io::BufReader;
+use std::path::Path;
+use std::path::PathBuf;
+use std::process;
+use std::process::Command;
+use std::sync::Once;
+
+// Synchronization primitive
+static INIT: Once = Once::new();
+
+#[derive(Debug)]
+struct Config {
+    language: Option<LANG>,
+    output_folder: String,
+}
+
+fn act_on_file(path: PathBuf, cfg: &Config) -> std::io::Result<()> {
+    // Open file
+    let source = if let Some(source) = read_file_with_eol(&path)? {
+        source
+    } else {
+        return Ok(());
+    };
+
+    // Guess programming language
+    let language = if let Some(language) = cfg.language {
+        language
+    } else if let Some(language) = guess_language(&source, &path).0 {
+        language
+    } else {
+        return Ok(());
+    };
+
+    // Build json file path
+    let file_path = Path::new(&cfg.output_folder)
+        .join(path.strip_prefix("./").unwrap())
+        .into_os_string()
+        .into_string()
+        .unwrap()
+        + ".json";
+
+    // Produce and compare metrics only if json file exists
+    if Path::new(&file_path).exists() {
+        // Get FuncSpace struct
+        let funcspace_struct = get_function_spaces(&language, source, &path, None).unwrap();
+
+        // Get Value struct from json file
+        let file = File::open(file_path)?;
+        let reader = BufReader::new(file);
+        let value_struct = serde_json::from_reader(reader)?;
+
+        // Compare the 2 structs
+        compare_structs(&funcspace_struct, &value_struct);
+    }
+
+    Ok(())
+}
+
+/// Compares a FuncSpace and a Value field by field
+fn compare_structs(funcspace: &FuncSpace, value: &serde_json::Value) {
+    let fsname = funcspace
+        .name
+        .as_deref()
+        .unwrap_or_default()
+        .replace("\r", "");
+    let name1 = Path::new(&fsname);
+    let name2 = Path::new(value["name"].as_str().unwrap_or_default());
+    assert_eq!(name1, name2);
+
+    let start_line1 = funcspace.start_line;
+    let start_line2 = value["start_line"].as_u64().unwrap() as usize;
+    assert_eq!(start_line1, start_line2);
+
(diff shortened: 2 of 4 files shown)


.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "tests/data/cpp/DeepSpeech"]
+	path = tests/data/cpp/DeepSpeech
+	url = https://github.com/mozilla/DeepSpeech.git
+[submodule "tests/data/rust/grcov"]
+	path = tests/data/rust/grcov
+	url = https://github.com/mozilla/grcov.git
```
marco-c, line 6: I think it'd be preferable to avoid submodules, and simply clone the repos at the time we have to run the tests. This way we don't force everyone to clone them even if they don't need to run the intergration tests, and the repository stays smaller.

marco-c:
Might be better to avoid storing all the expected JSON files in the repo, but run RCA before/after a change like we do for mozilla-central integration tests.
Another option would be to store all the JSON in a separate repository that we clone at the time of the integration tests, this way we don't need to store all these JSON files in this repository.

.gitmodules
marco-ballario: I will try to replace submodules with git clone commands at the beginning of my tests. In that way repositories are downloaded as first thing while running the tests (if not already present locally). Together with the repositories, I will also try to clone a specific version of RCA during my tests, so I can use it to produce the json files.

Thanks a lot for the suggestions!

marco-c: > I will also try to clone a specific version of RCA during my tests, so I can use it to produce the json files.

We can also store them in a separate repo, it might actually be better:
1) We avoid running RCA twice (before and after)
2) We have a history of how the values change over time with the evolution of RCA

marco-ballario: You're right, I can try that!
Where should I create the repository containing all the json files?

marco-c: You can create it in your personal GitHub account for now

marco-ballario: Michele gave me the permission to create the repository at https://github.com/SoftengPoliTo, so I will try to create it there.
Thanks for the support!

Luni-4 (approved):
Seems fine to me! Thank you!

src/tools.rs
```diff
@@ -232,7 +232,7 @@ pub fn guess_language<P: AsRef<Path>>(buf: &[u8], path: P) -> (Option<LANG>, Str
 pub(crate) fn remove_blank_lines(data: &mut Vec<u8>) {
     let count_trailing = data.iter().rev().take_while(|&c| *c == b'\n').count();
     if count_trailing > 0 {
-        data.truncate(data.len() - count_trailing + 1);
```
marco-c, line 235: Why is this change required?

src/concurrent_files.rs
```diff
@@ -92,6 +92,7 @@ where
         }
         if path.is_dir() {
             for entry in WalkDir::new(path)
+                .sort_by_file_name()
```
marco-c, line 95: Why is this change required?

tests/test.rs
```diff
@@ -0,0 +1,270 @@
...
+    compare_rca_output_with_files(
+        "v0.9.3",
+        "https://github.com/mozilla/DeepSpeech.git",
+        "./tests/repositories/DeepSpeech",
+        &["*.cc", "*.cpp", "*.h", "*.hh"],
+    );
+}
+
+#[test]
+fn test_grcov() {
```
marco-c, line 253: We could use the Rust standard library ("library" subfolder in https://github.com/rust-lang/rust, but you can just use the whole repo since you have the `.rs` glob) for testing Rust, this way we make sure we're good with latest Rust language features (they are definitely going to use them in the Rust standard library).

src/tools.rs
marco-ballario: During my tests I noticed that few DeepSpeech source files generated different json files in Windows and Linux. Thanks to @Luni-4 and @lu-zero help we determined that the cause of those differences were the end of line formats at the end of few downloaded source files. This line should fix this issue and allow the generation of the same json files both in Windows and in Linux.

src/concurrent_files.rs
marco-ballario: We found that file access order in a directory depends on the operating system policy. This line should define a unique deterministic exploration order valid for every operating system.

tests/test.rs
marco-ballario: Thank you for the review! Do you think I should substitute grcov tests with Rust standard library tests? Or I can leave grcov tests and simply add Rust standard library tests?

src/concurrent_files.rs
marco-c: Why do we need a deterministic exploration order?

tests/test.rs
marco-c: Both options work for me, you choose ;)

src/tools.rs
marco-c: Unless I'm mistaken, after your change this line is basically doing nothing (truncating `data` to its full length) and the `remove_blank_lines` function wouldn't actually remove blank lines.
@calixteman why was `remove_blank_lines` necessary?

Luni-4: I added that function to remove the blank lines at the end of a file, such that we don't consider them for the `Blank` metric counting. But on Windows new lines are identified by `/r/n`, so we have a different `data` length when we compare the source files parsed on Linux with the ones parsed on Windows. One solution could be leaving blank lines at the end of a file, as implemented in this PR (I agree with this approach because blank lines at the end of a file should be considered), otherwise we can find `\r\n` on Windows

src/concurrent_files.rs
marco-ballario: You are right. It seems this line is not strictly necessary for the tests to work. Source files can be read in any order since the json file path is deducted directly from the source file path. Comparisons for a single file will never be affected by this change. I will remove this change. Thank you for pointing this out!

src/tools.rs
marco-c: > One solution could be leaving blank lines at the end of a file, as implemented in this PR

So we could just remove the `remove_blank_lines` function entirely?

src/concurrent_files.rs
marco-c: No problem! I actually didn't know that it wasn't necessary, it's good I asked :)

src/tools.rs
Luni-4: Looking more carefully the code, `remove_blank_lines` removes blank lines when `count_trailing > 0` leaving a single new line at the end of a file. So my first solution is not feasible, the second one is correct instead

src/tools.rs
```diff
@@ -229,13 +229,17 @@ pub fn guess_language<P: AsRef<Path>>(buf: &[u8], path: P) -> (Option<LANG>, Str
     }
 }
 
+/// Replaces \n and \r ending characters with a single generic \n
 pub(crate) fn remove_blank_lines(data: &mut Vec<u8>) {
-    let count_trailing = data.iter().rev().take_while(|&c| *c == b'\n').count();
+    let count_trailing = data
+        .iter()
+        .rev()
+        .take_while(|&c| (*c == b'\n' || *c == b'\r'))
```
marco-c, line 237: This is going to count Windows newlines twice.
On Linux you would have `Ciao\nCiao\n\n`, and so count would be 2.
On Windows you would have `Ciao\r\nCiao\r\n\r\n`, and so count would be 4.

marco-c: Forget it, the code after is removing the characters and not the lines.

marco-c (approved):

PR #732, opened 2021-12-21
Change cli to produce output files in folders

marco-ballario:
This PR changes the cli file generation to produce files inside a directory structure.
With the previous file naming system, two files A and B:
A) a/b/c.cpp
B) a/b_c.cpp
Produced the same a_b_c.cpp.json file.
This PR solves this name ambiguity and helps to improve the integration tests.


Luni-4 (approved):
Thanks a lot!

PR #733, opened 2021-12-22
Add cargo udeps to CI

### LINKED ISSUE #726: Run 'cargo udeps' in CI
We can add it to pre-commit

Luni-4:
This PR adds `cargo udeps` to CI and fixes #726

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -41,6 +41,8 @@ tasks:
             - "-cx"
             - "curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3 get-pip.py &&
                rustup component add clippy rustfmt &&
+               rustup toolchain install nightly &&
+               curl -L https://github.com/est31/cargo-udeps/releases/download/v0.1.25/cargo-udeps-v0.1.25-x86_64-unknown-linux-gnu.tar.gz | tar xz -C /usr/local/cargo/bin --strip-components 2 &&
                git clone --quiet ${repository} &&
                cd rust-code-analysis &&
                git -c advice.detachedHead=false checkout ${head_rev} &&
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,6 +28,13 @@ repos:
         entry: cargo clippy --all-targets --all -- -D warnings
         pass_filenames: false
 
+      - id: udeps
+        name: udeps
+        language: system
+        files: '\.rs$'
+        entry: cargo +nightly udeps --all-targets
+        pass_filenames: false
+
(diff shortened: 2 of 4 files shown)


.taskcluster.yml
```diff
@@ -41,6 +41,8 @@ tasks:
             - "-cx"
             - "curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3 get-pip.py &&
                rustup component add clippy rustfmt &&
+               rustup toolchain install nightly &&
```
marco-c, line 44: Maybe we should use the `rustlang/rust:nightly` image instead of `rust:buster` so we don't have to install nightly in the task?

.taskcluster.yml
```diff
@@ -41,6 +41,8 @@ tasks:
...
+               curl -L https://github.com/est31/cargo-udeps/releases/download/v0.1.25/cargo-udeps-v0.1.25-x86_64-unknown-linux-gnu.tar.gz | tar xz -C /usr/local/cargo/bin --strip-components 2 &&
```
marco-c, line 45: What is this needed for?

.taskcluster.yml
Luni-4: Nightly clippy lints and fmt formatting are different from the ones found by stable version

.taskcluster.yml
Luni-4: To install `cargo udeps` fastly using the binary

.taskcluster.yml
marco-c: Is nightly required for cargo udeps?

Luni-4: Yes https://github.com/est31/cargo-udeps#usage

.taskcluster.yml
marco-c: Let's use cargo install instead, it's more future-proof.

Luni-4: The command took a while to run and slowed down the CI a lot. If we use the binary, we can pin a specific version avoiding CI breaking in the future also

marco-c: We can pin to a specific version anyway, by running `cargo install udeps --version 0.1.25` or simply adding it to the dev-dependencies in Cargo.toml (so we also benefit from Dependabot).
If it is slowing down the task too much we can keep it this way though.

marco-c (approved):

.taskcluster.yml
Luni-4: True about the `cargo install` command, just clarifying that we can pin to a specific version using the binary too

PR #738, opened 2021-12-30
Fix  comments at the end of function for python parsing

giovannitangredi:
For the python language the tree-sitter parser consider comments at the end of a function as outside of it instead of inside of the function.
For this reasons a added some code in spaces.rs so that the comments are inserted in the correct  FuncSpace.
Basically when a found a new function definition node I check all the next nodes to see if there are comments that should be inside the function space.
Link to PR: https://github.com/tree-sitter/tree-sitter-python/pull/143
Link to Issue: https://github.com/tree-sitter/tree-sitter-python/issues/113

### DIFF
--- a/src/spaces.rs
+++ b/src/spaces.rs
@@ -16,6 +16,7 @@ use crate::nargs::{self, NArgs};
 use crate::nom::{self, Nom};
 
 use crate::dump_metrics::*;
+use crate::langs::*;
 use crate::traits::*;
 
 /// The list of supported space kinds.
@@ -248,6 +249,7 @@ struct State<'a> {
 pub fn metrics<'a, T: ParserTrait>(parser: &'a T, path: &'a Path) -> Option<FuncSpace> {
     let code = parser.get_code();
     let node = parser.get_root();
+    let lang = parser.get_language();
     let mut cursor = node.object().walk();
     let mut stack = Vec::new();
     let mut children = Vec::new();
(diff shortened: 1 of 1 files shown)


marco-c:
Could you reformat the code with `cargo fmt`?

calixteman:
Is there an upstream bug about this issue ?

giovannitangredi:
It is a specific bug of the tree-sitter-python parser, There should be a pull request for resolving this issue in the tree-sitter-python github but they haven't still  reviewed and merged yet.

src/spaces.rs
```diff
@@ -293,7 +295,28 @@ pub fn metrics<'a, T: ParserTrait>(parser: &'a T, path: &'a Path) -> Option<Func
         cursor.reset(node.object());
         if cursor.goto_first_child() {
             loop {
-                children.push((Node::new(cursor.node()), new_level));
+                // FIX ME: This code fixes the python parser error for comments , when the tree-sitter-python parser will be fixed revert the changes
+                // Issue on the tree-sitter-python github: https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/25
+                // Link to PR resolving this issue : https://github.com/tree-sitter/tree-sitter-python/pull/143
+                if lang == LANG::Python && cursor.node().kind() == "function_definition" {
```
calixteman, line 301: Honestly, I don't like this patch...
I didn't dive into the detail of implementation here but the function (`metrics`) here is language agnostic and you're adding a check for a specific language (python) so we'll have this check for all the supported languages.

giovannitangredi: Ok, i'll see if i find a way to resolve this issue by not influencing the other supported languages

calixteman (commented):
After thinking about this PR... really I'm not super excited by the idea to fix some grammar issues in rca itself: grammar issues should be fixed in grammars themselves.
I don't say we mustn't do it: sometimes we can do it but the change is pretty small and very located in a language itself, but really we musn't add such a workaround in general file like this one.
If the upstream bug is really something which impacts the rca quality, then I way prefer using an updated grammar (like we do for the mozcpp one) else if the impact is small and limited then I guess we can wait for an upstream update.

Luni-4:
We could add a test that MUST fail due to the Python grammar error upstream through the `#[should_panic]` attribute. When the bug will be fixed upstream we are going to remove that attribute. In this way we are aware of the error without changing the rca code a lot. What do you think @calixteman?

calixteman:
> We could add a test that MUST fail, due to the Python grammar error upstream, using the `#[should_panic]` attribute. When the bug will be fixed upstream we are going to remove that attribute. In this way we are aware of the error without changing the rca code a lot. What do you think @calixteman?

I'm fine with such a "solution" (which is not a solution in fact).
And as said, if this bug has a real impact in real life and makes rca unusable or useless, we can workaround in adding our own grammar.

Issue #49, opened 2019-12-19
Remove operator stuff once the PR is merged


calixteman:
https://github.com/mozilla/rust-code-analysis/commit/e80db56c40f16305e39bf5b339c4928f8beca070

marco-c:
The PR was https://github.com/tree-sitter/tree-sitter-cpp/pull/73, and it was merged.

PR #757, opened 2022-01-13
Drop no longer needed C++ modification

### LINKED ISSUE #756: Drop no longer needed modification of the C++ grammar
We should be able to drop https://github.com/mozilla/rust-code-analysis/blob/1c3c663d99916a1abd009a86ddc30ef0514ee4e7/tree-sitter-mozcpp/grammar.js#L19-L38 as it should be part of upstream now.

We should try to drop it, regenerate grammar and check that there are no differences in metrics.

Luni-4:
This PR fixes #756 and #49


marco-c:
Did you try running the minimal tests stuff to see if there are differences?

marco-c:
Maybe we should update the mozcpp version to trigger the test on CI?

calixteman (approved):
It's useless now.

calixteman:
Normally it has been, but probably some other changes are responsible for this variations.

calixteman:
I check ts-cpp repo and my changed have been merged.
An other PR superseded them but the tests I wrote for my PR are of course still there:
https://github.com/tree-sitter/tree-sitter-cpp/blob/e8dcc9d2b404c542fd236ea5f7208f90be8a6e89/test/corpus/declarations.txt#L539-L543
So the operator bug is definitely fixed (the code Luni removed in this patch aimed to fix some issues with whitespaces in operator definition).
So from my pov, this PR is safe, or I missed something ;).

marco-c:
All right, sounds good then.

PR #786, opened 2022-02-10
Add Cyclomatic Complexity implementation for Java

marco-ballario:
This PR adds the Cyclomatic Complexity metrics implementation for the Java language. Two unit tests are added to verify the correct computation of the measures. Thanks in advance for the review.

### DIFF
--- a/src/metrics/cyclomatic.rs
+++ b/src/metrics/cyclomatic.rs
@@ -203,7 +203,19 @@ impl Cyclomatic for CppCode {
 
 impl Cyclomatic for PreprocCode {}
 impl Cyclomatic for CcommentCode {}
-impl Cyclomatic for JavaCode {}
+
+impl Cyclomatic for JavaCode {
+    fn compute(node: &Node, stats: &mut Stats) {
+        use Java::*;
+
+        match node.object().kind_id().into() {
+            If | For | While | Case | Catch | TernaryExpression | AMPAMP | PIPEPIPE => {
+                stats.cyclomatic += 1.;
+            }
+            _ => {}
+        }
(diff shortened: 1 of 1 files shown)


src/metrics/cyclomatic.rs
```diff
@@ -203,7 +203,19 @@ impl Cyclomatic for CppCode {
...
 impl Cyclomatic for PreprocCode {}
 impl Cyclomatic for CcommentCode {}
-impl Cyclomatic for JavaCode {}
+
+impl Cyclomatic for JavaCode {
+    fn compute(node: &Node, stats: &mut Stats) {
+        use Java::*;
+
+        match node.object().kind_id().into() {
+            If | For | While | Case | Catch | TernaryExpression | AMPAMP | PIPEPIPE => {
```
calixteman, line 212: It'd be nice in your tests to cover all the possibilities (there are not so much).

marco-ballario: Thank you for the advice! I extended my tests to cover all the different possibilities.

dburriss:
Nice good to see.

calixteman (approved):
LGTM. Thank you.

calixteman:
Unrelated to this change, but in looking to the code for js, I think we (I ??) missed to add operators like `??` and `?.` (maybe others ?): they're lazy evaluated like a `||` or a `&&`.
Oh and probably `||=`,  `&&=` are missing too.

Luni-4:
Yep, right, we forgot them. Could you please open an issue in order to remember them? Thank you @calixteman! :)

PR #806, opened 2022-03-20
fix: hanging server

tpraxl:
Since the update to actix-rt@2.3 (f413996a), the web server hasn't been
started, when we executed `cargo run -p rust-code-analysis-web`.

It froze once `actix_rt::System::new().run()?;` was called,
effectively preventing the execution of the code below.

By looking at the actix-rt examples and changing the code to be more
like the examples, I was able to fix this.
The webserver now behaves as expected.


marco-c (approved):
Thanks!

PR #807, opened 2022-03-20
Add Weighted Methods Per Class implementation for Java

marco-ballario:
This PR adds an implementation of the Weighted Method Per Class (WMC) object-oriented metric for Java.

The WMC for a class is the sum of the complexities of the methods defined in a class: 
https://www.researchgate.net/publication/3187649_Kemerer_CF_A_metric_suite_for_object_oriented_design_IEEE_Trans_Softw_Eng_206_476-493

Some unit tests are provided to cover the most common Java cases.
The code has been tested on 3 repositories with the following results:

For repositories 1 and 2 all the measures produced have the expected values.
For repository 3 the computation ends smoothly and never unexpectedly stops. The few files checked all provide the expected measures.

Thank you very much for the review.

### DIFF
--- a/src/output/dump_metrics.rs
+++ b/src/output/dump_metrics.rs
@@ -9,8 +9,9 @@ use crate::loc;
 use crate::mi;
 use crate::nargs;
 use crate::nom;
+use crate::wmc;
 
-use crate::spaces::{CodeMetrics, FuncSpace};
+use crate::spaces::{CodeMetrics, FuncSpace, SpaceKind};
 
 /// Dumps the metrics of a code.
 ///
@@ -104,7 +105,8 @@ fn dump_metrics(
     dump_halstead(&metrics.halstead, &prefix, false, stdout)?;
     dump_loc(&metrics.loc, &prefix, false, stdout)?;
     dump_nom(&metrics.nom, &prefix, false, stdout)?;
-    dump_mi(&metrics.mi, &prefix, true, stdout)
+    dump_mi(&metrics.mi, &prefix, false, stdout)?;
+    dump_wmc(&metrics.wmc, &prefix, true, stdout)
 }
 
 fn dump_cognitive(
@@ -291,6 +293,38 @@ fn dump_nexits(
 
     color!(stdout, White);
     writeln!(stdout, "{}", stats.exit())
+}
+
+fn dump_wmc(
+    stats: &wmc::Stats,
+    prefix: &str,
+    last: bool,
+    stdout: &mut StandardStreamLock,
+) -> std::io::Result<()> {
+    if !stats.is_not_class_or_unit() {
+        let (pref_child, pref) = if last { ("   ", "`- ") } else { ("|  ", "|- ") };
+
+        color!(stdout, Blue);
+        write!(stdout, "{}{}", prefix, pref)?;
+
+        color!(stdout, Green, true);
+        writeln!(stdout, "wmc")?;
+
+        let prefix = format!("{}{}", prefix, pref_child);
+        dump_value(
+            if stats.space_kind() == SpaceKind::Unit {
+                "wmc_total"
+            } else {
+                "wmc"
+            },
+            stats.wmc(),
+            &prefix,
+            true,
+            stdout,
+        )
+    } else {
+        Ok(())
+    }
 }
 
 fn dump_value(
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -207,7 +207,7 @@ impl Checker for JavaCode {
     mk_checker!(is_comment, Comment);
     mk_checker!(is_string, StringLiteral);
     mk_checker!(is_call, MethodInvocation);
-    mk_checker!(is_func, MethodDeclaration);
+    mk_checker!(is_func, MethodDeclaration, ConstructorDeclaration);
     mk_checker!(is_closure, LambdaExpression);
     mk_checker!(
         is_func_space,
--- a/src/getter.rs
+++ b/src/getter.rs
@@ -518,7 +518,7 @@ impl Getter for JavaCode {
         let typ = node.object().kind_id();
         match typ.into() {
             ClassDeclaration => SpaceKind::Class,
-            MethodDeclaration | LambdaExpression => SpaceKind::Function,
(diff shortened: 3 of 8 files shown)


rust-code-analysis-cli/src/main.rs
```diff
@@ -234,6 +252,7 @@ fn main() {
                 .long("metrics")
                 .short('m'),
         )
+        .arg(Arg::new("wmc").help("Compute WMC").long("wmc"))
```
calixteman, line 255: I know what `WMC` means thanks to the PR title but in two days I'll have forgotten.
So please, add something to explain what it is.

rust-code-analysis-cli/src/main.rs
```diff
@@ -101,6 +102,23 @@ fn act_on_file(path: PathBuf, cfg: &Config) -> std::io::Result<()> {
             let path = cfg.path.clone();
             action::<Metrics>(&language, source, &path, pr, cfg)
         }
+    } else if cfg.wmc {
+        if language == LANG::Java {
```
calixteman, line 106: Why is it Java only ?
What about js, c++, potentially C# ?

src/output/dump_metrics.rs
```diff
@@ -311,3 +312,101 @@ fn dump_value(
     color!(stdout, White);
     writeln!(stdout, "{}", val)
 }
+
+pub fn dump_root_wmc(file_classes: &FileClasses) -> std::io::Result<()> {
```
calixteman, line 316: I don't see any caller for this function.

calixteman: I've the feeling that you're re-inventing the space stuff...
I'd prefer you integrate that stuff in `dump_root`.
If the idea was to avoid to print something for a non-obj language, then we should probably add something to configure the dumper and then dump the wmc stuff when it's required.

src/wmc.rs
```diff
@@ -0,0 +1,566 @@
+// New Java switches are not correctly recognised by RCA, which returns some error nodes.
```
calixteman, line 1: nit: add a TODO.

calixteman, line 1: It'd be nice to have here and elsewhere in this file some urls linking to some useful documentation about wmc and the way to compute it.

src/wmc.rs
```diff
@@ -0,0 +1,566 @@
...
+            break;
+        } else {
+            let child_state = &mut state_stack.pop().unwrap();
+            let parent_state = state_stack.last_mut().unwrap();
+            merge_wmc(child_state, parent_state);
+        }
+    }
+}
+
+pub fn wmc_metrics<T: ParserTrait>(parser: &T, path: &Path) -> Option<FileClasses> {
```
calixteman, line 98: I've the impress that it's a modified copy/paste of the code in `spaces.rs::metrics`...
Isn't it possible to avoid that ?
I think we must try to improve `spaces.rs::metrics` to fit your needs or maybe you must change your way to think to fit `spaces.rs::metrics`...
Duplicating code is rarely a good idea... so tbh I'm not super enthusiastic about this PR.
But of course if you've a very very good rationale to do that, we can discuss ;)

calixteman (changes requested):
Overall, I've the impress you're plugging a new tool when I'd have imagined that you'll extend the `metrics` one.

rust-code-analysis-cli/src/main.rs
marco-ballario: I added the full metric name to the command so that it shows up when invoking `--help`

rust-code-analysis-cli/src/main.rs
marco-ballario: I didn't realize that line was useless. I removed it. Thank you for pointing it out!

src/wmc.rs
marco-ballario: I turned the comment into a FIXME and better explained that this is an issue already fixed upstream, in the tree-sitter-java GitHub repository

marco-ballario: I reported the same paper cited in the PR also in a comment

marco-ballario: Related to the dump_root comment above:
I introduced an enum to distinguish between dumping a FuncSpace or a FileClasses struct

src/wmc.rs
marco-ballario: Thank you very much for the code review and your comments.

The plan was to have an AST scan loop dedicated to all the object-oriented metrics computation.
In particular what I'm trying to do is to implement 3 of the 6 metrics defined in the CK metrics suite: WMC, DIT, and NOC.
WMC, DIT, and NOC are all metrics defined just for classes.
DIT (Depth of Inheritance Tree) and NOC (Number Of Children) are 2 metrics that focus on class inheritance.

WMC is quite similar to the Cyclomatic Complexity and I agree that it could be somehow implemented in `spaces.rs::metrics`.
However, to implement DIT and NOC I am next going to require a system to collect class data throughout different files before producing the DIT and NOC values for a single class.
Adding that system in metrics in my opinion will complicate a lot the function and will decrease a lot the performance.

Instead, I thought that having all 3 CK metrics using the same slimmer AST scan could simplify things:
1) I can use this faster AST scan to compute the object-oriented metrics that need to collect data from different files to produce values faster.
2) I can have all the 3 CK metrics computed in a simpler dedicated space so one can experiment more with them without any risks.
3) I can avoid printing out the WMC, DIT, and NOC values for spaces that are not classes since those metrics are not defined for non-class spaces.

Thank you for the opportunity to clarify and motivate my work.
Any additional advice and comment would be greatly appreciated and would help a lot to turn my work in the right direction.

marco-ballario:
Since DIT and NOC are too complex to implement at the moment, I modified the WMC to fit in the current metrics computation system and fixed all the issues related to the old implementation. Thank you in advance for the review!

Luni-4 (approved):
That's fine for me, all @calixteman comments have been addressed:
- No more duplicated code
- Good documentation with a link that shows how the metric works from a theoretical point of view
- A single function to `dump` the result

Thank you! :)

calixteman (approved):
LGTM. Thank you.

Issue #809, opened 2022-03-24
Add an Interface SpaceKind


dburriss:
I am opening this issue to have a discussion and record my findings. This may be something we don't want to do.

The idea is to add `SpaceKind::Interface` enum value so that interfaces show like classes (see image below).

This was kicked off from the conversation [here](https://github.com/mozilla/rust-code-analysis/pull/694#discussion_r832558571) with @calixteman .

**Scope and motivation**
This would be used in supported languages that make use `interface`s ie. Java, TypeScript (future C# and Go?). This makes more sense recently since, for good or ill, many languages support default implementations on interfaces. This would give other languages a little more parity with the Rust implementation which distinguishes `trait` and `impl` already.

**Effort and impact**
This appears to be a fairly trivial feature to add:
- [ ] Add `SpaceKind::Interface` enum
- [ ] Add `SpaceKind::Interface => "interface",` to the match in the `SpaceKind`  `Display` trait's `fmt` implementation

The following tasks are needed per language
- [ ] For language's `Getter`.`get_space_kind` return `SpaceKind::Interface` for correct grammar elements ie. `InterfaceDeclaration` in Java ie. `InterfaceDeclaration => SpaceKind::Interface`
- [ ] For the language's `Checker` implementation, add the grammar elements in the previous task to the `is_func_space` `mk_checker` macro arguments so the interface is included as a space for functions and metrics

I am not sure if this has much of a performance impact? I imagine not because of the type of change but I would need input or time to run tests.

**Result**
I implemented this for the Java lang and this is what it looks like. 
![image](https://user-images.githubusercontent.com/5071495/159846120-48a9e283-e915-4748-8452-965f30c0dbdb.png)

dburriss:
Some open questions for me on this:

- Do we even want to do this? Support this?
- Perf impact?
- Is this a breaking change? It does change the structure of output so are people depending on metrics being in only `Unit`, `Class`, or `Function`?

dburriss:
Closed by #810

PR #810, opened 2022-03-24
Interface SpaceKind

dburriss:
This is an implementation for #809 for adding the `SpaceKind::Interface` and enabling it as a function space in appropriate languages.

Supported languages that support `interface`:

- [x] Java
- [x] TypeScript
- [x] Tsx (including since different mechanism)

**Java**

```java
interface printable{  
  void print();  
}  
class Printer implements printable{  
  public void print(){System.out.println("Hello");}  
    
  public static void main(String args[]){  
    Printer obj = new Printer ();  
    obj.print();  
  }  
}  
```

unit: foo.java
- interface: printable
-- function: print
- class: Printer 
-- function:print

image

**TS**

```typescript
interface IPrintable { 
    print: ()=>void 
} 
 
class Printer implements IPrintable {  
    print ():void {return console.log("Hello");} 
} 
```

unit: foo.ts
- interface: IPrintable
-- function: print
- class: Printer
-- function:print

image

**TSX**

```tsx
interface IPrintable { 
    print: ()=>void 
}
```

unit: foo.tsx
- interface: IPrintable
-- function: print

image

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -209,7 +209,12 @@ impl Checker for JavaCode {
     mk_checker!(is_call, MethodInvocation);
     mk_checker!(is_func, MethodDeclaration);
     mk_checker!(is_closure,);
-    mk_checker!(is_func_space, Program, ClassDeclaration);
+    mk_checker!(
+        is_func_space,
+        Program,
+        ClassDeclaration,
+        InterfaceDeclaration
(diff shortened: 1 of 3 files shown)


dburriss:
@calixteman @marco-c Is this something we can merge in current state?

PR #811, opened 2022-03-25
Initial Halstead implementation for Java

dburriss:
- [x] Implement necessary functions
- [x] Passing test
- [x] Cleanup

Some references:

- [Java lang ref](https://docs.oracle.com/javase/specs/jls/se18/html/jls-3.html#jls-3.12)
- https://www.geeksforgeeks.org/software-engineering-halsteads-software-metrics/?msclkid=5e181114abef11ecbb03527e95a34828

### DIFF
--- a/src/getter.rs
+++ b/src/getter.rs
@@ -525,15 +525,6 @@ impl Getter for JavaCode {
         }
     }
 
-    // fn get_func_space_name<'a>(node: &Node, code: &'a [u8]) -> Option<&'a str> {
-    //     if let Some(name) = node.object().child_by_field_name("name") {
-    //         let code = &code[name.start_byte()..name.end_byte()];
-    //         std::str::from_utf8(code).ok()
-    //     } else {
-    //         Some("<anonymous>")
-    //     }
-    // }
-
     fn get_op_type(node: &Node) -> HalsteadType {
         use Java::*;
 
@@ -554,11 +545,6 @@ impl Getter for JavaCode {
         let typ = node.object().kind_id();
 
         match typ.into() {
-            // LPAREN | COMMA | STAR | GTGT | COLON | SEMI | Return | Break | Continue | If | Else
-            // | Switch | Case | Default | For | While | New | Try | Catch | Throw | Throws
-            // | Throws2 | EQ | EQEQ | AMP | AMPAMP | AMPEQ | LT | LTLT | LTEQ | LTLTEQ | BANGEQ
-            // | GTEQ | GTGTEQ | GTGTGT | GTGTGTEQ | PLUSEQ | BANG | STAREQ | SLASHEQ | PERCENTEQ
-            // | CARET | CARETEQ | LBRACE | RBRACE | QMARK | Instanceof 
             // Operator: function calls
             MethodInvocation
             // Operator: control flow
@@ -575,17 +561,17 @@ impl Getter for JavaCode {
             // type identifier
             | TypeIdentifier | IntegralType | FloatingPointType | BooleanType
             => {
-                println!("operator: {}", node.object().kind().to_string());
+                println!("operator: {}", node.object().kind());
                 HalsteadType::Operator
             },
 
             // Operands: variables, constants, literals
             Identifier | NullLiteral | ClassLiteral | StringLiteral | CharacterLiteral | HexIntegerLiteral | OctalIntegerLiteral | BinaryIntegerLiteral | DecimalIntegerLiteral | HexFloatingPointLiteral | DecimalFloatingPointLiteral  => {
-                println!("operand: {} ", node.object().kind().to_string());
+                println!("operand: {} ", node.object().kind());
                 HalsteadType::Operand
             },
             _ => {
-                println!("unknown: {} ", node.object().kind().to_string());
+                println!("unknown: {} ", node.object().kind());
                 HalsteadType::Unknown
             },
         }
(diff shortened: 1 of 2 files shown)


dburriss:
Test exmaple:

```java
public static void main(String args[]) {
  int a, b, c, avg;
  a = 5; b = 5; c = 5;
  avg = (a + b + c) / 3;
  MessageFormat.format(\"{0}\", avg);
}
```

I will get to this over the weekend and try to set my test up for expected numbers. 
@marco-c  who can explain to me the difference between `u_operators`, `operators`, `u_operands`, and `operands` in the Halstead tests?

Luni-4:
> Test exmaple:
> 
> ```java
> public static void main(String args[]) {
>   int a, b, c, avg;
>   a = 5; b = 5; c = 5;
>   avg = (a + b + c) / 3;
>   MessageFormat.format(\"{0}\", avg);
> }
> ```
> 
> I will get to this over the weekend and try to set my test up for expected numbers. @marco-c who can explain to me the difference between `u_operators`, `operators`, `u_operands`, and `operands` in the Halstead tests?

@dburriss 
`u_operators` are all distinct operators present in the code
`u_operands` are all unique operands present in the code 
`operators` are all occurrences of `u_operators` in the code
`operands` are all occurrences of `u_operands` in the code.

You can find what are operators and operands in these tests https://github.com/mozilla/rust-code-analysis/blob/master/src/ops.rs#L312 for different programming languages

dburriss:
@Luni-4 I have run into a weird issue.

When parsing the first line  `public static void main(String args[]){`

```
unknown: program 
unknown: local_variable_declaration 
unknown: modifiers 
unknown: public
unknown: static
operator: void_type
unknown: variable_declarator
operand: identifier
operator: ;
unknown: ERROR
unknown: formal_parameters
operator: (
unknown: formal_parameter
operator: type_identifier
operand: identifier
unknown: dimensions
operator: [
operator: ]
operator: )
unknown: block
```

Note the **operator: ;** and **unknown: ERROR** ... probably related. There is no **;** in the first line so this is throwing the count off.

The example is valid Java... I compiled it.

calixteman:
It's very likely a matter of adding `public class A {`and `}` around your function.

dburriss:
> It's very likely a matter of adding `public class A {`and `}` around your function.

Thanks @calixteman
That did the trick. It actually crossed my mind but I had never seen ERROR in the LoC parsing and most of those tests don't have class. Anyway, I learned something. And I am getting the  numbers I expect.

@Luni-4 I included a lot of comments and links, etc. in the code explaining what and why I am including different things. I didn't find definitive rules about what is included or excluded.

image

dburriss:
@Luni-4 

I looked at *ops.rs* and added a test but having some problems.

Copying the existing tests quickly and filling in from the list above (or the code), I get the following error:

```
thread 'ops::tests::java_ops' panicked at 'assertion failed: `(left == right)`
  left: `["()", ")", "+", ",", ".", "/", ";", "=", "[]", "]", "integral_type", "method_invocation", "type_identifier", "void_type", "{}", "}"]`,
 right: `["()", "+", ",", "/", ";", "=", "String", "[]", "format", "int", "void", "{}"]`', src\ops.rs:280:9
```

A few questions questions:

1. Do you know why the list for Java is showing the grammar type instead of the value ie. 'integral_type` instead of `int` as it does for the other languages
2. Does the order of operators/operands matter in the assert?
3. Are these `println` supposed to be here? (hard to believe but they are not mine :) ) https://github.com/mozilla/rust-code-analysis/blob/master/src/ops.rs#L92

dburriss:
@Luni-4 this is failing on the *ops.rs* file you mentioned. Not something I introduced. The linter has an opinion on something. I can make the change in this PR but I'm not sure how this got in in the first place. New rule maybe?

Log: https://community-tc.services.mozilla.com/tasks/WqxZrlJ2TIKZMbpDibX5Jg/runs/0/logs/public/logs/live_backing.log

dburriss:
Shuffled things around a bit and moved ops check to #849
Hopefully, it is ok to merge those 2 separately.
@calixteman @Luni-4 ok to review with this scope?

halstead eg

src/getter.rs
```diff
@@ -524,4 +524,42 @@ impl Getter for JavaCode {
...
+            | EQEQ | LTEQ | GTEQ | BANGEQ | AMPAMP | PIPEPIPE | PLUSPLUS | DASHDASH
+            | PLUS | DASH | STAR | SLASH | AMP | PIPE | CARET | PERCENT| LTLT | GTGT | GTGTGT
+            | PLUSEQ | DASHEQ | STAREQ | SLASHEQ | AMPEQ | PIPEEQ | CARETEQ | PERCENTEQ | LTLTEQ | GTGTEQ | GTGTGTEQ
+            // type identifier
+            | TypeIdentifier | IntegralType | FloatingPointType | BooleanType
+            => {
+                HalsteadType::Operator
+            },
+            // Operands: variables, constants, literals
+            Identifier | NullLiteral | ClassLiteral | StringLiteral | CharacterLiteral | HexIntegerLiteral | OctalIntegerLiteral | BinaryIntegerLiteral | DecimalIntegerLiteral | HexFloatingPointLiteral | DecimalFloatingPointLiteral  => {
```
Luni-4, line 555: Can we split this long line one more lines?

Luni-4 (changes requested):
Sorry @dburriss for the delay, yeah, it's fine splitting up the two problems.

Just a final request and we can land this PR

src/getter.rs
dburriss: I did find it weird the formatter didn't do that. Done 👍

Luni-4 (approved):
Thank you @dburriss! Fine for me!

Issue #789, opened 2022-02-12
Add support for C-Sharp


dburriss:
I imagine we would use [tree-sitter-c-sharp](https://github.com/tree-sitter/tree-sitter-c-sharp).

Since the Java metrics are growing I would like to add C# too. I imagine the implementations should be close to identical.

marco-c:
Good idea!

dburriss:
So folding in lessons learned from #694 I am starting with some research and definitions. Thankfully C# has a fairly clear definition of a statement as well as a comprehensive list of the different types:

[csharp statements](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/statements)

A cursory look through seems to indicate, unsurprisingly, that it is much the same as Java. I will raise it if I find any weird difference.

Should I create an issue per metric to track this or link PRs here?

dburriss:
So bumping into my first challenge here.
On this repo currently tree-sitter version dependency is`tree-sitter = ">= 0.19, < 0.21"`.  
The source Cargo.toml for tree-sitter-c-sharp is `tree-sitter = "0.19"`.  
On [crates.io](https://docs.rs/crate/tree-sitter-c-sharp/0.16.1) the tree-sitter version is 0.17.

I guess this is the reason I get *expected struct `tree_sitter::Language`, found a different struct `tree_sitter::Language* when trying to add `tree-sitter-c-sharp` as a dependency?

I assume the fix here is to ask @dcreager, the owner of the crate, if it can be updated?

dcreager:
Just published `tree-sitter-c-sharp` 0.19.1, which should unblock you if you're using `tree-sitter` 0.19.

If you're using `tree-sitter` 0.20, then I think you'll end up getting the same error, until https://github.com/tree-sitter/tree-sitter-c-sharp/pull/219 (which I just opened) is merged.  Once that gets a :+1: review I'll merge it and cut a new release of `tree-sitter-c-sharp` with the more inclusive `tree-sitter` dependency.

dburriss:
Thanks so much @dcreager 🙏
I'll test it out tomorrow but I think it should work.

marco-c:
> Should I create an issue per metric to track this or link PRs here?

Let's just use this one, maybe you can edit the first comment to add a list of metrics with checkmarks.

dburriss:
After adding `tree_sitter_c_sharp` to the `nums` crate I get the follwong compile error:

```
[...]

Issue #674, opened 2021-09-16
Update tree-sitter and grammars to version 0.20


marco-c:

dburriss:
Based on my issues adding `tree-sitter-c-sharp` I thought I would do a little look at the feasibility of rolling forward to 0.20.

- [ ] tree-sitter-java (version = "0.19.0" | tree-sitter = "0.19" | [PR](https://github.com/tree-sitter/tree-sitter-java/pull/86))
- [x] tree-sitter-typescript (version = "0.20.0" | tree-sitter = ">= 0.19, < 0.21")
- [x] tree-sitter-javascript (version = "0.20.0" | tree-sitter = ">= 0.19, < 0.21")
- [ ] tree-sitter-python (version = "0.19.0" | tree-sitter = ">= 0.19, < 0.21" | [PR](https://github.com/tree-sitter/tree-sitter-python/pull/120))
- [x] tree-sitter-rust (version = "0.20.1" | tree-sitter = "0.20" )

For c-sharp there is a [PR](https://github.com/tree-sitter/tree-sitter-c-sharp/pull/219) mentioned in #789 by the maintainer.

marco-c:
The PRs are being ignored by the maintainers, we could use forks and specify dependencies as git repositories (https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories).
For example, for tree-sitter-python, we could specify `tree-sitter-python = { git = "https://github.com/Luni-4/tree-sitter-python", branch = "update-0.20" }`.

dburriss:
@marco-c I was looking at #812 and was wondering if we have a final action plan here?
That PR looks like it is uncovering some good bugs uncovered by the upgrade (or some that were there already?).

I did notice Python points to its tree-sitter master and the other to a fork from @BrainBuzzer that is a work in progress.
In the merge, this would point to something more stable than master though.

I do want to ask though, what does 0.20 give us? Were there important security fixes, perf, features, that the project benefited from? Couldn't we settle on 0.19 until the situation improves with dependencies? This only makes sense if we think the situation CAN be improved. If it isn't something any of us can influence, then the forking strategy is the way to go.

PR #812, opened 2022-03-31
Update all dependencies to use tree-sitter v0.20

BrainBuzzer:
Updated all dependencies to use tree-sitter v0.20. This PR includes:
1. Regenerated grammar rules for all languages
2. Java has changed the `Comment` type to `LineComment` and `BlockComment`
3. Fixes for JavaScript tests (added more about this in the commit description)

There are some issues with tests for the rust repository. It seems that unlike the previous version where functions in macros were detected and counted towards metrics are not detected anymore due to changes in how the tree-sitter parses code. The following metrics are broken depending on macros used in rust code:
- [ ] Cyclomatic
- [ ] Cognitive
- [x] Halstead
- [ ] MI
- [ ] NOM

Note - Python is using the GitHub repo's master branch because v0.19.1 from crates.io for `tree-sitter-python` was not up to date and was producing more errors than expected. It might be a good idea to fork and tag the repo to the latest commit.


BrainBuzzer:
Related discussions: #674 #789

BrainBuzzer:
After investigating in the test cases, found following:

1. ExpressionStatement was declared before every Statement, leading to almost double `lloc`.
2. The previous version of the tree-sitter used to give false positives, which led to a higher number of operators and operands. This in turn, was also affecting the metrics related to halstead. An example of this is in the file [library/proc_macro/src/lib.rs#L994](https://github.com/rust-lang/rust/blob/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/proc_macro/src/lib.rs#L994), you can see that in the comment there is a dash. In the previous version, this dash was being counted as an operator. Perhaps it's time to update `rca-output` repo once this PR manages to produce proper results.
3. It also seems that the cyclomatic complexity for rust code is incorrect. The cyclomatic compute function for `RustCode` utilizes two questionable node types. `For` and `QMARK`. Say there's a line as follows:
```rust
impl<T: ?Sized> UnwindSafe for Mutex<T> {}
```
  Do we want to count the cyclomatic complexity of the above code as 2?

Investigating further problems with Cognitive, NOM, and MI. cc @marco-c @marco-ballario

BrainBuzzer:
After checking through the rest of the metrics, I'm 99% certain that the previous metrics were giving a lot of false positives, and that's why most of them have an error margin of ~1-2%. This PR otherwise seems ready to me for merging.

tjbell:
I was curious if there was any movement on this pull request? It looks like the main tree-sitter-java grammar was just updated to 20.0.0.

Luni-4:
@tjbell 

Yeah, the grammars was just updated but not on `crates.io` which is still at `0.19`. We are trying to come up with a solution here if you are interested in the conversation https://github.com/tree-sitter/tree-sitter/discussions/1768

PR #822, opened 2022-04-12
Adds exit metrics

dburriss:
- [x] Implement *nom* as needed for this metric (i think looking at htis: https://github.com/mozilla/rust-code-analysis/blob/0cc9bcdf51464777ccc1a0cf1fab9883cca6dc7e/src/spaces.rs#L178)
- [x] Implementation
- [x] Tests (some done during development but probably should add another 1 or 2)

before exit metric

after exit metric

### DIFF
--- a/src/metrics/exit.rs
+++ b/src/metrics/exit.rs
@@ -172,9 +172,16 @@ impl Exit for CppCode {
     }
 }
 
+impl Exit for JavaCode {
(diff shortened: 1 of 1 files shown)


dburriss:
@marco assuming all checks pass this is ready for review.

marco-c (approved):
Thanks!

PR #823, opened 2022-04-13
Java nom metrics

dburriss:
Adds `LambdaExpression` to the `Checker` `is_closure` and adds a test for Java nom metric.

image

### DIFF
--- a/src/metrics/nom.rs
+++ b/src/metrics/nom.rs
@@ -123,6 +123,7 @@ impl Stats {
     /// Returns the average number of function definitions over all spaces
     #[inline(always)]
     pub fn functions_average(&self) -> f64 {
+        println!("SPACE {}", self.space_count);
         self.functions_sum() / self.space_count as f64
     }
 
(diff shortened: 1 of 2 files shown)


marco-c (dismissed):
Makes sense to me. Could you also add a smaller test like `cpp_nom` in addition to the one you already added?

dburriss:
> Makes sense to me. Could you also add a smaller test like `cpp_nom` in addition to the one you already added?

Yeah... turns out lambdas in Java are really ugly :)

src/metrics/nom.rs
```diff
@@ -123,6 +123,7 @@ impl Stats {
     /// Returns the average number of function definitions over all spaces
     #[inline(always)]
     pub fn functions_average(&self) -> f64 {
+        println!("SPACE {}", self.space_count);
```
marco-c, line 126: Please remove the debugging statement ;)

dburriss: Bah! Forgot about it. Done.

marco-c (approved):
Thanks!

PR #836, opened 2022-05-06
Add ABC metric implementation for Java

marco-ballario:
This PR adds an implementation of the ABC metric for the Java language.

The ABC metric measures the size of a source code by counting the number of Assignments (A), Branches (B), and Conditions (C). The metric defines an ABC score as a vector of three elements (```<A,B,C>```). The ABC score can be represented by its individual components (A, B, and C) or by the magnitude of the vector (```|<A,B,C>| = sqrt(A^2 + B^2 + C^2)```). The metric was introduced by Jerry Fitzpatrick in 1997 to overcome the drawbacks of the LOC.

Source: *Applying the ABC Metric to C, C++, and Java* (https://www.softwarerenovation.com/Articles.aspx)

Some unit tests are provided to cover the most common Java cases. The code has been tested on these 3 repositories with the following results:

For repositories 1 and 2 all the measures have the expected values. For repository 3 the computation ends smoothly and never unexpectedly stops. The few JSON files checked all provide the expected measures.

Thank you in advance for the review.

### DIFF
--- a/src/metrics/abc.rs
+++ b/src/metrics/abc.rs
@@ -1,0 +1,1070 @@
+use serde::ser::{SerializeStruct, Serializer};
+use serde::Serialize;
+use std::fmt;
+
+use crate::checker::Checker;
+use crate::node::Node;
+use crate::*;
+
+/// The `ABC` metric.
+///
+/// The `ABC` metric measures the size of a source code by counting
+/// the number of Assignments (`A`), Branches (`B`) and Conditions (`C`).
+/// The metric defines an ABC score as a vector of three elements (`<A,B,C>`).
+/// The ABC score can be represented by its individual components (`A`, `B` and `C`)
+/// or by the magnitude of the vector (`|<A,B,C>| = sqrt(A^2 + B^2 + C^2)`).
+///
+/// Official paper and definition:
+///
+/// Fitzpatrick, Jerry (1997). "Applying the ABC metric to C, C++ and Java". C++ Report.
+/// https://www.softwarerenovation.com/Articles.aspx
+#[derive(Debug, Clone)]
+pub struct Stats {
+    assignments: f64,
+    assignments_sum: f64,
+    assignments_min: f64,
+    assignments_max: f64,
+    branches: f64,
+    branches_sum: f64,
+    branches_min: f64,
+    branches_max: f64,
+    conditions: f64,
+    conditions_sum: f64,
+    conditions_min: f64,
+    conditions_max: f64,
+    space_count: usize,
+    declaration: Vec<DeclKind>,
+}
+
+#[derive(Debug, Clone)]
+pub enum DeclKind {
+    Var,
+    Const,
+}
+
+impl Default for Stats {
+    fn default() -> Self {
+        Self {
+            assignments: 0.,
+            assignments_sum: 0.,
+            assignments_min: f64::MAX,
+            assignments_max: 0.,
+            branches: 0.,
+            branches_sum: 0.,
+            branches_min: f64::MAX,
+            branches_max: 0.,
+            conditions: 0.,
+            conditions_sum: 0.,
+            conditions_min: f64::MAX,
+            conditions_max: 0.,
+            space_count: 1,
+            declaration: Vec::new(),
+        }
+    }
+}
+
+impl Serialize for Stats {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut st = serializer.serialize_struct("abc", 13)?;
+        st.serialize_field("assignments", &self.assignments_sum())?;
+        st.serialize_field("branches", &self.branches_sum())?;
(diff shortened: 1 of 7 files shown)


src/metrics/abc.rs
```diff
@@ -0,0 +1,803 @@
...
+/// https://www.softwarerenovation.com/Articles.aspx (page 5)
+///
+/// "These are cases where a single variable/field/value is treated as a boolean value.
+/// Examples include if (x) and return !ready."
+///
+/// - GMetrics (a Groovy source code analyzer):
+/// https://dx42.github.io/gmetrics/metrics/AbcMetric.html
+/// - GMetrics test examples:
+/// https://github.com/dx42/gmetrics/blob/master/src/test/groovy/org/gmetrics/metric/abc/AbcMetric_MethodTest.groovy
+fn count_java_unary_conditions(node: &Node) -> Option<(f64, bool)> {
```
Luni-4, line 274: Why does this function return an `Option` value? It always returns the `Some` at the end because there are `break`s inside the `while let` loop. We can remove it

src/metrics/abc.rs
```diff
@@ -0,0 +1,803 @@
...
+                res += 1.;
+            }
+
+            // Ignores parenthesis and not operators and looks inside them
+            while let ParenthesizedExpression | UnaryExpression = condition_kind {
+                // Sets flag if a not operator is found
+                // This is useful to know if a variable or a value returned by a method is proved to be a boolean
+                // e.g. `return (!x);`
+                if matches!(condition_kind, UnaryExpression) {
+                    is_bool = true;
```
Luni-4, line 300: We can replace the `if` with: `is_bool = matches!(condition_kind, UnaryExpression)`

src/metrics/abc.rs
```diff
@@ -0,0 +1,803 @@
...
+                        }
+                        _ => {}
+                    }
+                }
+            }
+            // Counts unary conditions inside for statements
+            ForStatement => {
+                // Position of the condition if the for counter has already been declared outside the for
+                // e.g. `for ( i=0; ... ; ... ) {}`
+                if let Some(condition) = node.object().child(3) {
```
Luni-4, line 408: Perhaps we can explicitly state that `3` is related to the comment above saying: *3 magic number represents this situation...*

src/metrics/abc.rs
```diff
@@ -0,0 +1,803 @@
...
+                    match condition.kind_id().into() {
+                        SEMI => {
+                            // Position of the condition if the for counter is declared inside the for
+                            // e.g. `for ( int i=0; ... ; ... ) {}`
+                            if let Some(cond) = node.object().child(4) {
```
Luni-4, line 413: Same here: *4 magic number represents this situation...*

src/metrics/abc.rs
```diff
@@ -0,0 +1,803 @@
...
+            }
+            SEMI => {
+                if let Some(DeclKind::Const | DeclKind::Var) = stats.declaration.last() {
+                    stats.declaration.clear();
+                }
+            }
+            EQ => {
+                // Excludes constant declarations
+                if let Some(decl) = stats.declaration.last() {
+                    if let DeclKind::Var = decl {
```
Luni-4, line 362: Can we use a `matches!` function here?

src/metrics/abc.rs
```diff
@@ -0,0 +1,803 @@
...
+                if let Some(parent) = node.object().parent() {
+                    if !matches!(parent.kind_id().into(), TypeArguments) {
+                        stats.conditions += 1.;
+                    }
+                }
+            }
+            // Counts unary conditions of elements separated by boolean operators
+            AMPAMP | PIPEPIPE => {
+                if let Some(parent) = node.object().parent() {
+                    if let Some((res, _)) = count_java_unary_conditions(&Node::new(parent)) {
```
Luni-4, line 386: Instead of using a `res`, can we pass `stats.conditions` as `count_java_unary_conditions` argument and update its value inside that function? In this way we remove a series of duplicated code. Same for all the conditions above

Luni-4 (changes requested):
Thank you for your PR, @marco-ballario! 

Just a series of comments to clean up the code a bit and improve the documentation

src/metrics/abc.rs
```diff
@@ -0,0 +1,1037 @@
...
+
+    let mut node = container_node.object();
+    let mut node_kind = node.kind_id().into();
+    let is_boolean_container = matches!(
+        node.parent().unwrap().kind_id().into(),
+        BinaryExpression | IfStatement | WhileStatement | DoStatement | ForStatement
+    );
+    let mut is_boolean_content = is_boolean_container;
+    let mut is_parenthesised_exp;
+    let mut is_not_operator;
```
Luni-4, line 268: We can remove these two lines and replace the ones inside the loop in this way:

```
let  is_parenthesised_exp = ..
let is_not_operator = ...
```

Luni-4 (approved):
Seems good to me now! Thanks for your work @marco-ballario!

PR #850, opened 2022-05-30
Cognitive complexity metric for Java

dburriss:
Last metric for #359 

- [x] Implementation
- [x] All tests pass
- [x] Branch up-to-date
- [x] Cleanup

Tests and API have changed a fair bit since I started work on this so tracking some additional tasks:

- [x] update API usage to accommodate changes
- [x] redo tests with Insta
- [x] figure out why average not working (something to do with space functions I think)


marco-c:
@dburriss any plans to finish this?

dburriss:
Hi @marco-c 
Sorry was job hunting and then ramping up so didn't have much energy for this. I'll try dive into this again tomorrow morning now that things have settled.

Luni-4:
Hello @dburriss, thank you for your hard work! 

I think that if you insert test methods inside a class, you would get the correct average value

Luni-4:
> Hi @Luni-4 Thanks for the suggestion. I'll give that a try.

It worked!

dburriss:
@marco-c  @Luni-4 It looks like this is finally ready for a second pair of eyes.

src/metrics/cognitive.rs
```diff
@@ -1696,4 +1734,189 @@ mod tests {
...
+        check_metrics::<JavaParser>(
+            "class X {
+                public static void print(Boolean a, Boolean b, Boolean c, Boolean d){  
+                    if(a && b){ // +2
+                        System.out.println(\"test1\");
+                    }
+                    if(c && d){ // +2
+                        System.out.println(\"test2\");
+                    }
+                }}",
```
Luni-4, line 1822: Class terminator on a new line. Let's do the same even for the other classes

dburriss:
@Luni-4 thanks for the review. I have made the requested improvements.

src/metrics/cognitive.rs
```diff
@@ -1696,4 +1733,197 @@ mod tests {
...
+                );
+            },
+        );
+    }
+
+    #[test]
+    fn java_switch_statement() {
+        check_metrics::<JavaParser>(
+            "class X {
+              public static void print(boolean a, boolean b, boolean c, boolean d){  // +1
```
Luni-4, line 1847: We should remove `+1` here, otherwise the test function is wrong

src/metrics/cognitive.rs
```diff
@@ -1696,4 +1733,197 @@ mod tests {
...
+                );
+            },
+        );
+    }
+
+    #[test]
+    fn java_switch_expression() {
+        check_metrics::<JavaParser>(
+            "class X {
+              public static void print(boolean a, boolean b, boolean c, boolean d){  // +1
```
Luni-4, line 1880: We should remove `+1` here, otherwise the test function is wrong

src/metrics/cognitive.rs
```diff
@@ -1696,4 +1733,197 @@ mod tests {
...
+                );
+            },
+        );
+    }
+
+    #[test]
+    fn java_not_booleans() {
+        check_metrics::<JavaParser>(
+            "class X {
+              public static void print(boolean a, boolean b, boolean c, boolean d){  // +1
```
Luni-4, line 1908: We should remove `+1` here, otherwise the test function is wrong

src/metrics/cognitive.rs
```diff
@@ -1696,4 +1733,197 @@ mod tests {
...
+                if (a && !(b && c)) { // +3 (+1 &&, +1 &&)
```
Luni-4, line 1909: We should change the comment otherwise the sum is not correct. As it is written right now, the sum is 2 not 3.

src/metrics/cognitive.rs
```diff
@@ -1696,4 +1733,197 @@ mod tests {
...
+            "###
+            );
+        });
+    }
+
+    #[test]
+    fn java_single_branch_function() {
+        check_metrics::<JavaParser>(
+            "class X {
+                public static void print(Boolean a){
```
Luni-4, line 1758: We should replace `Boolean` with `boolean`

Luni-4 (changes requested):
Thank you @dburriss! Just simple things again

src/metrics/cognitive.rs
dburriss: Is there a reason for the change by the way? Both are valid. Change made though as I missed this one from previous request. Thanks.

src/metrics/cognitive.rs
dburriss: Is there a reason for the change by the way? Both are valid. Change made though as I missed this one from previous request. Thanks.

Luni-4: Right, forgot to explain the reasons of that choice. `Boolean` is the class that wraps a `boolean` type and it is not automatically imported, so when I had passed the codes to a static analyzer it returned me errors on our test functions. Replacing directly with the type fixed the error and our code remains correct because we use them only for comparisons

src/metrics/cognitive.rs
```diff
@@ -1696,4 +1733,197 @@ mod tests {
...
+            },
+        );
+    }
+
+    #[test]
+    fn java_not_booleans() {
+        check_metrics::<JavaParser>(
+            "class X {
+              public static void print(boolean a, boolean b, boolean c, boolean d){
+                if (a && !(b && c)) { // +1 (+1 &&, +1 &&)
```
Luni-4, line 1909: We can use this notation, so it is clearer why the value is 3: https://github.com/mozilla/rust-code-analysis/blob/4167e9ba17641816021c839fd1e45f3666714592/src/metrics/cognitive.rs#L951

Luni-4 (changes requested):
Ok, last change and we are ready to merge in!

src/metrics/cognitive.rs
dburriss: Sorry, @Luni-4 I miss understood what you were asking for so I changed this incorrectly. That should be correct comment now.

dburriss:
@marco-c the task expired before a worker was made available for the checks. Should I push another commit to trigger or there an issue (and trying again could make it worse).

src/metrics/cognitive.rs
dburriss: @Luni-4 looks like the checks timed out waiting for an agent

marco-c:
I retriggered it manually and it passed.

Luni-4 (dismissed):
It's fine for me! Thanks a lot @dburriss!

dburriss:
@marco-c the steps that had not run passed but the rest were in an exception state due to timeout exceeded. Triggered a check with a commit.
@Luni-4 I updated a comment to get the check to run. You may want to check it since I've got those wrong before :p

Luni-4 (approved):
@dburriss Your last commit is fine as well for me! Thank you!

PR #857, opened 2022-06-15
Add NPM metric implementation for Java

marco-ballario:
This PR adds an implementation of the Number of Public Methods (NPM) metric for the Java language.

The NPM counts the number of public methods declared in a class: https://pdepend.org/documentation/software-metrics/number-of-public-methods.html

The implementation also calculates NPM averages over NOM to know the percentage of public methods in a class. Those averages are an adaptation of the COA (Classified Operation Accessibility) metric for not classified methods: https://sci-hub.hkvisa.net/10.1109/QSIC.2009.11

Some unit tests are provided to cover the most common Java cases. The code has been tested on these 3 repositories with the following results:

For repositories 1 and 2 all the measures have the expected values. For repository 3 the computation ends smoothly and never unexpectedly stops. The few JSON files checked all provide the expected measures.

Thank you in advance for the review!

### DIFF
--- a/src/metrics/mod.rs
+++ b/src/metrics/mod.rs
@@ -7,4 +7,5 @@ pub mod loc;
 pub mod mi;
 pub mod nargs;
 pub mod nom;
+pub mod npm;
 pub mod wmc;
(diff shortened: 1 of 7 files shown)


Luni-4 (changes requested):
Thanks for your work! 

I guess we should try to re-implement the `skip_serializing` part with a new approach, let's see if it's feasibile

PR #881, opened 2022-07-04
Update WMC to include Java interfaces and empty classes

marco-ballario:
This PR applies few changes to the WMC metric:
- Fixes a bug that prevented the visualization and serialization of the metric for empty Java classes
- Includes Java interfaces in the computation
- Adds few unit tests for Java interfaces
- Simplify the overall metric and its dump


Luni-4 (approved):
Seems good to me, thank you!

PR #889, opened 2022-07-04
Use new stable total_cmp function to compare floats

Luni-4:
Since Rust 1.62, it is possible to compare float with the [total_cmp](https://doc.rust-lang.org/std/primitive.f64.html#method.total_cmp) function which makes use of a standard comparison method from IEEE

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -318,9 +318,8 @@ macro_rules! check_metrics {
 
             $(
                 $(
-                    let value: f64 = $true_float_value;
-                    if !value.is_nan() {
-                        assert!((func_space.metrics.$metric.$func_float() - value).abs() < f64::EPSILON);
+                    if !($true_float_value as f64).is_nan() {
+                        assert!(func_space.metrics.$metric.$func_float().total_cmp(&$true_float_value) == std::cmp::Ordering::Equal);
                     } else {
                         assert!(func_space.metrics.$metric.$func_float().is_nan());
                     }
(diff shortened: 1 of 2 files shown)


src/macros.rs
```diff
@@ -318,12 +318,11 @@ macro_rules! check_metrics {
...
             $(
                 $(
-                    let value: f64 = $true_float_value;
-                    if !value.is_nan() {
-                        assert!((func_space.metrics.$metric.$func_float() - value).abs() < f64::EPSILON);
+                    assert!(if ($true_float_value as f64).is_nan() {
+                        func_space.metrics.$metric.$func_float().is_nan()
                     } else {
-                        assert!(func_space.metrics.$metric.$func_float().is_nan());
-                    }
```
marco-c, line 326: Could you keep the old structure and just change the contents of the first if branch?

marco-c (changes requested):
CI appears to be failing

src/macros.rs
Luni-4: Ok, I preferred this solution to have a compact way into a single `assert`, reducing duplicated code

Luni-4:
> CI appears to be failing

It seems a spurious errore, try to re-run it, it should work

src/macros.rs
marco-c: I understand, but it makes the diff more complex and is unrelated to this change. You could do it in a separate commit or PR.

marco-c (approved):

PR #890, opened 2022-07-06
Update WMC to include Java files without class or interface declarations

marco-ballario:
This PR applies few changes to the WMC metric:
- Fixes a bug that prevented the metric visualization and serialization for Java files without class or interface declarations
- Moves the merge operation into a new function


Luni-4 (approved):
Seems good to me and simplifies the code a bit, thank you!

PR #912, opened 2022-08-18
Make id and root public in AstNode

Squadrick:
I'm trying to use `rust-code-analysis` as a library. I can't access the node in the result of `AstCallback`. The hack-y way is to serialize and deserialize to void the access check, but this should be public.


Squadrick:
@Luni-4 @marco-c Please review.

Luni-4 (approved):
Seems fine to me, thank you!

Squadrick:
Some clippy errors are blocking this merge. Should be fixed by https://github.com/mozilla/rust-code-analysis/pull/913.

PR #916, opened 2022-08-24
Replace single WMC field with class and interface sums

marco-ballario:
This PR splits the single WMC value into two fields, corresponding to classes and interfaces.
It also introduces class and interface sums and a total field.
Moreover this PR modifies the WMC computation so that the space kind field is updated at most once.
Serialization, dump and unit tests have been updated accordingly.

The code has been tested on 2 small repositories ([Simple-Java-Calculator](https://github.com/pH-7/Simple-Java-Calculator) and [Notepad](https://github.com/buiducnhat/Notepad)) and 1 large repository ([netbeans](https://github.com/apache/netbeans)).
For the small repositories all the measures have the expected values.
For the large repository the computation ends smoothly and never unexpectedly stops.

Thanks in advance for the review!


Luni-4 (approved):
It's fine for me, thank you!

PR #938, opened 2022-10-11
Fix the version for the `rust-code-analysis` crate

suhaibmujahid:
This should avoid installing incompatible versions for `rust-code-analysis-cli` and `rust-code-analysis-web` when a newer version of `rust-code-analysis` is released.

For example, when Cargo installs `rust-code-analysis-web` v0.0.23, it uses the latest version of `rust-code-analysis` (i.e., v0.0.24) as a dependency, which includes compatibility breakage. 

This will not solve the problem for `rust-code-analysis-web` v0.0.23 since it is already released, but this will avoid having the same problem again.


Luni-4 (approved):
Good catch, thank you!

PR #953, opened 2022-12-06
Use insta

lu-zero:
To use it when the snapshots aren't there yet
```
cargo install cargo-insta
cargo insta test --no-ignore --accept
```


Luni-4:
Blocked by https://github.com/mitsuhiko/insta/issues/314

PR #958, opened 2022-12-12
Add Cargo.lock only to the release job

Luni-4:
`Cargo.lock` causes many dependencies issues during a development and more in general causes to waste 2x the development time because it makes difficult the rebase operation.

For these reasons, we have added it only to the release job


Luni-4:
Blocked by #954

marco-c (dismissed):

marco-c:
There is one case where this is really useful: being notified of tree-sitter updates so we can check that there are no regressions with our integration tests.

Currently, in Cargo.toml we have `tree-sitter = ">= 0.20, < 0.21"` and so dependabot only checks Cargo.lock for new patch versions. Here's an example PR from dependabot: #922.

Maybe the best solution would be to remove Cargo.lock but force a specific tree-sitter version in Cargo.toml, so we can still rely on dependabot to notify us of new versions and we can run integration tests before upgrading.

marco-c (approved):

PR #967, opened 2022-12-14
Use insta for the integration tests

lu-zero:
The rust repo is removed since it will be replaced with a stable codebase for the rust case.

### DIFF
--- a/src/spaces.rs
+++ b/src/spaces.rs
@@ -164,8 +164,10 @@ impl FuncSpace {
                 node.object().end_position().row + 1,
             ),
         };
+
         Self {
-            name: T::get_func_space_name(node, code).map(|name| name.to_string()),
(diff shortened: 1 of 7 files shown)


lu-zero:
The windows tests are still problematic but would be good to investigate them at a later time.

src/spaces.rs
```diff
@@ -164,8 +164,10 @@ impl FuncSpace {
                 node.object().end_position().row + 1,
             ),
         };
+
         Self {
-            name: T::get_func_space_name(node, code).map(|name| name.to_string()),
+            name: T::get_func_space_name(node, code)
+                .map(|name| name.split_whitespace().collect::<Vec<_>>().join(" ")),
```
marco-c, line 170: Why was this necessary?

marco-c (approved):

PR #970, opened 2022-12-16
Check each grammar on independent CI tasks

Luni-4:
Each grammar will be checked in search of regression in a independent CI task. This mechanism is going to reduce the overall computation time since tasks are run in parallel.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -154,7 +154,7 @@ tasks:
           owner: cdenizet@mozilla.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
-      - taskId: {$eval: as_slugid("check_tree_sitter_crates")}
+      - taskId: {$eval: as_slugid("check_tree_sitter_mozcpp")}
         dependencies:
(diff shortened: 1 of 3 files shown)


Luni-4:
> @Luni-4 I'd add other grammars too, we run these tasks seldomly so it's OK to pay the price to be more confident about our changes.

Done

.taskcluster.yml
```diff
@@ -171,19 +171,217 @@ tasks:
...
+              type: file
+        scopes:
+          - "docker-worker:cache:rust-code-analysis-mozilla-central-repository"
+        metadata:
+          name: rust-code-analysis check tree-sitter-python
+          description: rust-code-analysis check tree-sitter-python grammar
+          owner: cdenizet@mozilla.com
+          source: ${repository}/raw/${head_rev}/.taskcluster.yml
+
+      - taskId: {$eval: as_slugid("check_tree_sitter_javascript")}
```
marco-c, line 322: We already have a task for mozjs, so we don't need a task for javascript

marco-c (approved):

PR #1000, opened 2023-01-27
Env variables should be strings

lotas:
Soon schemas would be strict about task payload and would not allow non-string env variables


matt-boris (approved):
LGTM - affecting PR [here](https://github.com/taskcluster/taskcluster/pull/6013).

marco-c (approved):

marco-c:
Thanks!

PR #1014, opened 2023-03-30
Replace color macro with simpler functions

Luni-4:
This PR replaces `color` macro with simpler functions


calixteman:
Why do you want to do that ?

marco-c (approved):

PR #1015, opened 2023-03-30
Remove mk checker macro

Luni-4:
This PR removes `mk_checker` macro for these reasons:
- State a checker trait clearly for each language. Trait functions which are not implemented are overridden by the base trait implementation, so it is not necessary to use a macro without any argument as in precedence
- `mk_checker` macro is confusing to read and handle

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -81,7 +81,21 @@ macro_rules! is_js_func_and_closure_checker {
 }
 
 pub trait Checker {
-    fn is_comment(node: &Node) -> bool;
+    fn is_comment(_: &Node) -> bool;
+    fn is_useful_comment(_: &Node, _: &[u8]) -> bool;
+    fn is_func_space(_: &Node) -> bool;
+    fn is_func(_: &Node) -> bool;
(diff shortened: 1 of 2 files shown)


marco-c:
I would prefer not to have a default implementation, to make sure the implementators of the trait have to specify everything and don't miss anything by mistake

marco-c:
@Luni-4 could you rebase?

src/checker.rs
```diff
@@ -174,6 +200,52 @@ impl Checker for CppCode {
...
     }
 
+    fn is_func_space(node: &Node) -> bool {
+        matches!(
+            node.object().kind_id().into(),
+            Cpp::TranslationUnit
+                | Cpp::FunctionDefinition
+                | Cpp::FunctionDefinition2
+                | Cpp::FunctionDefinition3
+                | Cpp::FunctionDefinition4
```
marco-c, line 210: This was not present before and is present now, so it is a behavior change. Could you add a test to cover it?

marco-c (approved):

PR #1016, opened 2023-03-31
Simplify is_primitive function in checker

Luni-4:
This PR simplifies the `is_primitive` function removing superfluous matches, replacing them with simpler comparisons, and useless attributes

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -179,7 +179,7 @@ impl Checker for CppCode {
 
     #[inline(always)]
     fn is_primitive(id: u16) -> bool {
-        matches!(id.into(), Cpp::PrimitiveType)
+        id == Cpp::PrimitiveType
     }
 }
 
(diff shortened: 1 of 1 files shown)


marco-c:
Why is the inline attribute useless?

Luni-4:
From [here](https://doc.rust-lang.org/nightly/reference/attributes/codegen.html?highlight=inline): When applied to a function in a trait, they apply only to that function when used as a default function for a trait implementation and not to all trait implementations. The attributes have no effect on a trait function without a body.

marco-c:
I think they are useless in functions defined in a trait, if that function is not overridden by an implementation.
In this case, for example for `is_primitive` in `impl Checker for CppCode`, the attribute is applied to a function in an implementation of a trait, so it is considered by the compiled in this case.

Luni-4:
> I think they are useless in functions defined in a trait, if that function is not overridden by an implementation. In this case, for example for `is_primitive` in `impl Checker for CppCode`, the attribute is applied to a function in an implementation of a trait, so it is considered by the compiled in this case.

Ok, seems legit, I'm going to leave them then

marco-c (approved):

PR #1017, opened 2023-03-31
Remove else if macro

Luni-4:
This PR removes `mk_else_if` macro because it was difficult to read and it is easier to write a comparison function for each language inside the dedicated trait


marco-c:
LGTM, why are you removing the inline attributes? Could you keep them and remove them in a separate PR if it is needed?

marco-c (approved):

PR #1022, opened 2023-04-05
Improve structs format in web crate

Luni-4:
This PR improves structs format in web crate and move it alongside their effective implementation


marco-c:
Why is `pub struct WebCommentCallback;` better than `pub struct WebCommentCallback {}'?

Luni-4:
Just following what is written on the [Rust Book](https://doc.rust-lang.org/book/ch05-01-defining-structs.html#unit-like-structs-without-any-fields) about unit structs

marco-c (approved):

PR #1029, opened 2023-04-06
Update test submodules

Luni-4:
This PR updates test submodules to last version


marco-c:
Why there are >4000 files changed in the reference files but way fewer in the reference repositories?

Luni-4:
> Why there are >4000 files changed in the reference files but way fewer in the reference repositories?

Because I removed JSONs related to the old integration tests force pushing the submodule, but if you click on `rca-output` repository you can see that points to the right repository

marco-c:
OK, next time it'd be better to do one step at a time and avoid force pushing, so we don't lose history of what we did.

marco-c (approved):

PR #1032, opened 2023-04-06
Abstract tree-sitter functions

Luni-4:
This PR extends the `Node` module adding more functions to interact with nodes without having to use `tree-sitter` functions directly. It also adds a `Cursor` struct that provides some functions to visit a tree and its nodes.

The goal of this PR are:
- Create internal APIs to interact with the AST
- Make `tree-sitter` completely transparent to a developer, making more simple a possible future replacement
- Make the code more modular

### DIFF
--- a/src/alterator.rs
+++ b/src/alterator.rs
@@ -16,11 +16,11 @@ where
             "".to_string()
         };
         if span {
-            let spos = node.object().start_position();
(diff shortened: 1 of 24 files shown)


marco-c (approved):
Thanks!

Issue #864, opened 2022-06-17
Create a sink thread to write down files


Luni-4:
It would be better that all files are written by a sink thread that collects all produced data and then writes them down

PR #1036, opened 2023-04-14
Write down files on main thread for cli

Luni-4:
This PR fixes part of #864 

Instead of leaving threads in writing down metrics and operators, use the main thread to perform that operation. In this way a burden of opening and closing files is completely handled by the main thread and not by each receiver thread


marco-c:
> Instead of leaving threads in writing down metrics and operators, use the main thread to perform that operation. In this way a burden of opening and closing files is completely handled by the main thread and not by each receiver thread

What is the benefit of doing this?

Luni-4:
In my opinion, I can see two main benefits:
- The first one is related to usability. If we save spaces produced by each thread in a vector, then we can aggregate them in some way or compute transversal metrics using their information. So saving data first, and then writes it down somewhere, can be useful in that perspective, because we can also customize more the output.
- The second one is related to performances. We do not know how each operating system handles file writing, there could be some operating systems which have some bottlenecks when I/O operations are performed on threads, especially a writing operation. So assigning all of that to the main thread allows the underlying operating system to entirely handle the output task, organizing the entire flow as it prefers.

marco-c:
>     * The first one is related to usability. If we save spaces produced by each thread in a vector, then we can aggregate them in some way or compute transversal metrics using their information. So saving data first, and then writes it down somewhere, can be useful in that perspective, because we can also customize more the output.

This works for me if we have a specific use case in mind, otherwise this seems like a premature optimization and we should only do it when the necessity actually arises.

>     * The second one is related to performances. We do not know how each operating system handles file writing, there could be some operating systems which have some bottlenecks when I/O operations are performed on threads, especially a writing operation. So assigning all of that to the main thread allows the underlying operating system to entirely handle the output task, organizing the entire flow as it prefers.

I'm not sure we will see any benefits, actually I see more chances for regressions than improvements as we serialize I/O operations. Can you run some benchmarks to make sure?

Luni-4:
> This works for me if we have a specific use case in mind, otherwise this seems like a premature optimization and we should only do it when the necessity actually arises.

We would like to add an option that produces a single output file containing all metrics of each input source file. That would help to reduce the effort in analyzing repositories data when you have to aggregate information.

> I'm not sure we will see any benefits, actually I see more chances for regressions than improvements as we serialize I/O operations. Can you run some benchmarks to make sure?

I run `rust-code-analysis-cli`, `release` version, on `mozilla-central` using at frist `master` branch and then this branch. I've used `time` to record time and memory usage.

My system specification:

```
       memory
          description: System memory
          size: 16GiB
       cpu
          product: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
          size: 1938MHz
          capacity: 5GHz
          width: 64 bits
```

Comparison:

**master 13453728cdfaa109bb261e3b25b4f2b527724952**

```
 Command being timed: "./rust-code-analysis-cli -m -O json -o /home/out-mozilla/ -p /home/mozilla-central/"
	User time (seconds): 1250.27
	System time (seconds): 2018.10
	Percent of CPU this job got: 970%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 5:36.76
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 4106548
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 26
	Minor (reclaiming a frame) page faults: 2273722
	Voluntary context switches: 350173
	Involuntary context switches: 73652
	Swaps: 0
	File system inputs: 5302184
	File system outputs: 6184520
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
```

**sink c1d7fb8f43f710feb5e9620a4e87904f8e3c6417**

```
Command being timed: "./rust-code-analysis-cli -m -O json -o /home/out-mozilla/ -p /home/mozilla-central/"
	User time (seconds): 1071.20
	System time (seconds): 668.04
	Percent of CPU this job got: 205%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 14:07.37
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 6276840
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 13
	Minor (reclaiming a frame) page faults: 2402231
	Voluntary context switches: 299067
	Involuntary context switches: 12983
	Swaps: 0
	File system inputs: 3576456
	File system outputs: 6184512
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
```

The great amount of time (and memory too) makes this solution infeasible.

We can try to define a unique file handler where each thread writes on. That should reduce time and memory usage from a conceptual point of view.

marco-c:
> We would like to add an option that produces a single output file containing all metrics of each input source file. That would help to reduce the effort in analyzing repositories data when you have to aggregate information.

This should be feasible independently. E.g. each thread can write a separate file and then at the end you merge them into a single one.

Luni-4:
> > We would like to add an option that produces a single output file containing all metrics of each input source file. That would help to reduce the effort in analyzing repositories data when you have to aggregate information.
> 
> This should be feasible independently. E.g. each thread can write a separate file and then at the end you merge them into a single one.

Yes, but in that case we should open every file again, we can obtain the same adding a cli option that tells thread to write on a specific file. It should be more feasible

PR #1041, opened 2023-04-21
Document web module

### LINKED ISSUE #383: Document web module
- [ ] Document `web` module
- [ ] Add a Readme to `web` module

Luni-4:
This PR fixes #383.

Unfortunately doc tests on binary cannot still be executed because of https://github.com/rust-lang/rust/issues/50784#issuecomment-898572571, but we can create them the same waiting for the feature to be implemented

Otherwise we can transform this module in a lib, with a binary associated, in this way we can document all APIs and at the same time having an executable

### DIFF
--- a/rust-code-analysis-web/Cargo.toml
+++ b/rust-code-analysis-web/Cargo.toml
@@ -8,9 +8,6 @@ keywords = ["metrics"]
 description = "Run a web service to compute and export code metrics"
 license = "MPL-2.0"
 
-[[bin]]
-name = "rust-code-analysis-web"
-
 [dependencies]
 actix-rt = "^2.6"
 actix-web = "^4.2"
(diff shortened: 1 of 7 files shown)


Luni-4:
I finished documenting the web module, converted it in a library in order to add more tests and doc-tests, and also improved the `server` test

rust-code-analysis-web/src/bin/rust-code-analysis-web.rs
```diff
@@ -1,15 +1,8 @@
-// After adding new fields for min and max in the json (server.rs line 630) this error arose:
```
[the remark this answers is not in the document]
marco-c: Ah you moved it.

rust-code-analysis-web/Cargo.toml
```diff
@@ -8,9 +8,6 @@ keywords = ["metrics"]
 description = "Run a web service to compute and export code metrics"
 license = "MPL-2.0"
 
-[[bin]]
-name = "rust-code-analysis-web"
```
marco-c, line 12: Will the release still contain a "web" binary if we do this?

Luni-4: Yes, because the name of the binary is the same filename of the Rust file contained in the `bin` directory

marco-c (approved):

PR #1042, opened 2023-04-21
Make some traits private

Luni-4:
This PR makes `CodeMetricsT` and `TSLanguage` traits private, since they are only used inside the `lib` crate to simplify language constructions, thus they should not be used outside the main lib

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -36,10 +36,13 @@ pub trait Callback {
     fn call<T: ParserTrait>(cfg: Self::Cfg, parser: &T) -> Self::Res;
 }
 
-#[doc(hidden)]
-pub trait CodeMetricsT:
-    Cognitive + Cyclomatic + Exit + Halstead + NArgs + Loc + Nom + Mi + Wmc + Abc + Npm + Npa
-{
+pub(crate) mod private {
+    use super::*;
+
+    pub trait CodeMetricsT:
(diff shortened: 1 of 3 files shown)


src/traits.rs
```diff
@@ -36,19 +36,21 @@ pub trait Callback {
     fn call<T: ParserTrait>(cfg: Self::Cfg, parser: &T) -> Self::Res;
 }
 
-#[doc(hidden)]
-pub trait CodeMetricsT:
-    Cognitive + Cyclomatic + Exit + Halstead + NArgs + Loc + Nom + Mi + Wmc + Abc + Npm + Npa
-{
-}
+pub(crate) mod private {
```
marco-c, line 39: Let's use `_private` so it's even clearer

src/traits.rs
```diff
@@ -36,19 +36,21 @@ pub trait Callback {
     fn call<T: ParserTrait>(cfg: Self::Cfg, parser: &T) -> Self::Res;
 }
 
-#[doc(hidden)]
```
marco-c, line 39: I'd say we should keep them hidden from the docs

Luni-4: They do not appear in documentation the same, so inserting the attribute is useless. They only appear in the traits required by `Parser` struct, but we need to refactor the code more heavily to change that outocome

marco-c (approved):

PR #1045, opened 2023-04-28
Refactor command line format enumerator

Luni-4:
This PR refactors the command line format enumerator splitting up the code in functions which are more comprehensible at first glance and reduces useless references passage. It leaves the main functionalities untouched.

### DIFF
--- a/rust-code-analysis-cli/src/formats.rs
+++ b/rust-code-analysis-cli/src/formats.rs
@@ -57,11 +57,13 @@ impl FromStr for Format {
     }
 }
 
+#[inline(always)]
 fn print_on_stdout(content: String) {
     writeln!(std::io::stdout().lock(), "{content}").unwrap();
 }
 
 trait WriteOnStdout {
+    #[inline(always)]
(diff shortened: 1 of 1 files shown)


rust-code-analysis-cli/src/formats.rs
```diff
@@ -21,100 +21,118 @@ impl Format {
...
-                    }
-                }
-                Format::Toml => {
-                    let toml_data = if pretty {
-                        toml::to_string_pretty(&space).unwrap()
-                    } else {
-                        toml::to_string(&space).unwrap()
-                    };
-                    format_file.write_all(toml_data.as_bytes())
+            self.dump_stdout(space, pretty)
```
marco-c, line 33: It's a pity that there is so much code duplication between these two functions.

marco-c (approved):

rust-code-analysis-cli/src/formats.rs
Luni-4: Yep, you're right, I'm going to simplify these functions in this PR

marco-c (approved):

PR #1046, opened 2023-04-28
Make metric traits public

Luni-4:
This PR makes all metric traits public. Code changes in this code have the following reasons:
- `CodeMetricsT` was only an aggregator of traits, better to declare metrics explicitly in `Parser` struct and `ParserTrait` removing a boilerplate layer
- Do not hide all code metric traits in documentation
- Add a macro that implements in a fake way unimplemented code metrics traits

We have decided to make all trait public in order to allow external users to compute metrics on single nodes since all arguments of the `compute` function are already public

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -222,7 +222,6 @@ macro_rules! mk_code {
     ( $( ($camel:ident, $code:ident, $parser:ident, $name:ident, $docname:expr) ),* ) => {
         $(
             pub struct $code { _guard: (), }
-            impl _private::CodeMetricsT for $code { }
 
             impl _private::TSLanguage for $code {
                 type BaseLang = $camel;
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -13,6 +13,56 @@ macro_rules! get_language {
     };
 }
 
+macro_rules! implement_metric_trait {
+    (Abc, $($code:ident),+) => (
+        $(
+           impl Abc for $code {
+               fn compute(_node: &Node, _stats: &mut Stats) {}
+           }
+        )+
+    );
+    (Cognitive, $($code:ident),+) => (
+        $(
+           impl Cognitive for $code {
+               fn compute(_node: &Node, _stats: &mut Stats, _nesting_map: &mut FxHashMap<usize, (usize, usize, usize)>,) {}
+           }
+        )+
+    );
+    (Halstead, $($code:ident),+) => (
+        $(
+           impl Halstead for $code {
+               fn compute<'a>(_node: &Node<'a>, _code: &'a [u8], _halstead_maps: &mut HalsteadMaps<'a>) {}
+           }
+        )+
+    );
+    (Loc, $($code:ident),+) => (
+        $(
+           impl Loc for $code {
+               fn compute(_node: &Node, _stats: &mut Stats, _is_func_space: bool, _is_unit: bool) {}
+           }
+        )+
+    );
+    (Wmc, $($code:ident),+) => (
+        $(
+           impl Wmc for $code {
+               fn compute(_space_kind: SpaceKind, _cyclomatic: &cyclomatic::Stats, _stats: &mut Stats) {}
+           }
+        )+
+    );
+    ([$trait:ident], $($code:ident),+) => (
+        $(
+           impl $trait for $code {}
+        )+
+    );
+    ($trait:ident, $($code:ident),+) => (
+        $(
+           impl $trait for $code {
+               fn compute(_node: &Node, _stats: &mut Stats) {}
+           }
+        )+
+    )
+}
+
 macro_rules! mk_enum {
     ( $( $camel:ident, $description:expr ),* ) => {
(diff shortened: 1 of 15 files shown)


marco-c:
I'm OK with these changes, except the last two commits that do `Add a macro that implements in a fake way unimplemented code metrics traits`. What is the benefit of this?

Luni-4:
Without that macro, we should add `{}` to the trait definition not to have a compiler error on the implementations. In this way instead:
- We have a clean definition with a correct documentation (since traits are now public)
- If a code language does not implement a trait, the compiler immediately raises an error

Luni-4:
> Are you sure we don't achieve the same two results if we stop at commit [232f502](https://github.com/mozilla/rust-code-analysis/commit/232f502b60017f8eeea4a799434425470cc92e66)? Isn't `impl Abc for PythonCode {}` already giving us an error if a language doesn't implement a trait?

Yes, we already have an error but because of [this line](https://github.com/mozilla/rust-code-analysis/pull/1046/commits/e93be2a5a5a8ae376e134b4f46e459ebef99782d#diff-4a04259da480a6b794a2e947e4cc03eff4d1aa9330836f5b91cac68c5398193fR103), since we require to have `Abc` trait implemented for `PythonCode` to allow `Parser<PythonCode>`.
To better explain my position: using a trait with a default implementation `does` not create a clear documentation and leads to lose the distinctions among metric implementations for each language

marco-c:
Could you show me the difference in generated documentation? I want to understand the result better before accepting the second part of the changes (I'm happy to merge the first part right away if you create that as a separate PR).

Luni-4:
Yep, loading the two documentations to show the main differences

Documentation without the second part

Screenshot from 2023-05-05 08-47-00

Documentation with the second part

Screenshot from 2023-05-05 08-49-40

The main differences are:

- No empty implementation within the trait, just the function signature
- No unused variable warnings due to the empty implementation, so the removal of `_` from function arguments

For what concerns the error I was talking about instead:

Commenting `PythonCode` implementation on `master` for `abc` metric

```
error[E0277]: the trait bound `langs::PythonCode: metrics::abc::Abc` is not satisfied
   --> src/macros.rs:225:45
    |
225 |               impl _private::CodeMetricsT for $code { }
    |                                               ^^^^^ the trait `metrics::abc::Abc` is not implemented for `langs::PythonCode`
    |
   ::: src/langs.rs:8:1
    |
8   | / mk_langs!(
9   | |     // 1) Name for enum
10  | |     // 2) Language description
11  | |     // 3) Display name
...   |
126 | |     )
127 | | );
    | |_- in this macro invocation
    |
    = help: the following other types implement trait `metrics::abc::Abc`:
              langs::CcommentCode
              langs::CppCode
              langs::JavaCode
              langs::JavascriptCode
              langs::KotlinCode
              langs::MozjsCode
              langs::PreprocCode
              langs::RustCode
            and 2 others
note: required by a bound in `CodeMetricsT`
   --> src/traits.rs:43:83
    |
42  |     pub trait CodeMetricsT:
    |               ------------ required by a bound in this trait
43  |         Cognitive + Cyclomatic + Exit + Halstead + NArgs + Loc + Nom + Mi + Wmc + Abc + Npm + Npa
    |                                                                                   ^^^ required by this bound in `CodeMetricsT`
    = note: this error originates in the macro `mk_code` which comes from the expansion of the macro `mk_langs` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
error: could not compile `rust-code-analysis` due to previous error
```

Commenting `PythonCode` implementation on this PR for `abc` metric

```
error[E0277]: the trait bound `langs::PythonCode: Abc` is not satisfied
   --> src/macros.rs:207:38
    |
207 |                           let parser = $parser::new(source, &path, pr);
    |                                        ^ the trait `Abc` is not implemented for `langs::PythonCode`
    |
   ::: src/langs.rs:8:1
    |
8   | / mk_langs!(
9   | |     // 1) Name for enum
10  | |     // 2) Language description
11  | |     // 3) Display name
...   |
126 | |     )
127 | | );
    | |_- in this macro invocation
    |
    = help: the following other types implement trait `Abc`:
              langs::CcommentCode
              langs::CppCode
              langs::JavaCode
              langs::JavascriptCode
              langs::KotlinCode
              langs::MozjsCode
              langs::PreprocCode
              langs::RustCode
            and 2 others
note: required by a bound in `parser::Parser`
   --> src/parser.rs:34:11
    |
29  | pub struct Parser<
    |            ------ required by a bound in this struct
...
34  |         + Abc
    |           ^^^ required by this bound in `Parser`
    = note: this error originates in the macro `mk_action` which comes from the expansion of the macro `mk_langs` (in Nightly builds, run with -Z macro-backtrace for more info)
```

The main difference is that the error is now associated to the `Parser` struct  and not to the `CodeMetricsT` trait implementation

marco-c (approved):
All right, I'm convinced, thanks for the additional details!

PR #1047, opened 2023-05-12
Make tree-sitter API for language retrieval private

Luni-4:
Make `tree-sitter` language API for language retrieval, only used privately to create a parser, private.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -102,6 +102,16 @@ macro_rules! mk_impl_lang {
                     )*
                 }
             }
+
+            // Returns a tree-sitter language.
+            // This function is only used to construct a parser.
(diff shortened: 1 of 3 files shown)


src/macros.rs
```diff
@@ -102,6 +102,14 @@ macro_rules! mk_impl_lang {
                     )*
                 }
             }
+
+            pub(crate) fn get_ts_language(&self) -> Language {
```
marco-c, line 106: Could you add docs like for the function above?

Luni-4: With "docs", are you referring to documentation or comments that explain this function?

marco-c: Just comments, like the function just above this one

marco-c (approved):

PR #1078, opened 2023-10-23
Added QualifiedIdentifier kinds; Closes #1055

### LINKED ISSUE #1055: empty function name for cpp
When I run analysis for cxx file, it is unable to identify class member function name.

For example, the member function
```
void CuiConfigFunctionListBox::ClearAll()
/*  Description
    Deletes all entries in the FunctionListBox, all UserData and all
    possibly existing MacroInfo.
*/
{
    sal_uInt16 nCount = aArr.size();
    for ( sal_uInt16 i=0; i<nCount; ++i )
    {
        SfxGroupInfo_Impl *pData = aArr[i].get();

        if ( pData->nKind == SfxCfgKind::FUNCTION_SCRIPT )
        {
            OUString* pScriptURI = static_cast<OUString*>(pData->pObject);
            delete pScriptURI;
        }

        if ( pData->nKind == SfxCfgKind::GROUP_SCRIPTCONTAINER )
        {
            XInterface* xi = static_cast<XInterface *>(pData->pObject);
            if (xi != nullptr)
            {
                xi->release();
            }
        }
    }

    aArr.clear();
    m_xTreeView->clear();
}
```
its output is
```
{"name":null,"start_line":310,"end_line":339,"kind":"function","spaces":[],"metrics":{"nargs":{"total_functions":0.0,"total_closures":0.0,"average_functions":0.0,"average_closures":0.0,"total":0.0,"average":0.0,"functions_min":0.0,"functions_max":0.0,"closures_min":0.0,"closures_max":0.0},"nexits":{"sum":0.0,"average":0.0,"min":0.0,"max":0.0},"cognitive":{"sum":8.0,"average":8.0,"min":8.0,"max":8.0},"cyclomatic":{"sum":5.0,"average":5.0,"min":5.0,"max":5.0},"halstead":{"n1":18.0,"N1":65.0,"n2":23.0,"N2":43.0,"length":108.0,"estimated_program_length":179.1005750152729,"purity_ratio":1.6583386575488233,"vocabulary":41.0,"volume":578.6156164987531,"difficulty":16.82608695652174,"level":0.05943152454780362,"effort":9735.836677609454,"time":540.8798154227475,"bugs":0.1519827383557585},"loc":{"sloc":30.0,"ploc":23.0,"lloc":12.0,"cloc":4.0,"blank":3.0,"sloc_average":30.0,"ploc_average":23.0,"lloc_average":12.0,"cloc_average":4.0,"blank_average":3.0,"sloc_min":30.0,"sloc_max":30.0,"cloc_min":4.0,"cloc_max":4.0,"ploc_min":23.0,"ploc_max":23.0,"lloc_min":12.0,"lloc_max":12.0,"blank_min":3.0,"blank_max":3.0},"nom":{"functions":1.0,"closures":0.0,"functions_average":1.0,"closures_average":0.0,"total":1.0,"average":1.0,"functions_min":1.0,"functions_max":1.0,"closures_min":0.0,"closures_max":0.0},"mi":{"mi_original":81.67528282898678,"mi_sei":69.4405024202566,"mi_visual_studio":47.763323291805136},"abc":{"assignments":0.0,"branches":0.0,"conditions":0.0,"magnitude":0.0,"assignments_average":0.0,"branches_average":0.0,"conditions_average":0.0,"assignments_min":0.0,"assignments_max":0.0,"branches_min":0.0,"branches_max":0.0,"conditions_min":0.0,"conditions_max":0.0}}}
```
where the `name` is `null`.

It seems that it is only able to identify names of those functions without `namespace::` on the left. Is this behavior expected? Is it better to extract names for class member functions?

michaelvanstraten:
There currently is an issue with this tool where when a scope resolution operator is use, aka the `::` in `Foo::bar`, in a function name the resulting in the name metric of that function will be `null`. This is due to the function name having the `Cpp::QualifiedIdentifier2` kind. I've added the `QualifiedIdentifier`, `QualifiedIdentifier2`, `QualifiedIdentifier3`, `QualifiedIdentifier4` to be on the safe side.

### DIFF
--- a/src/getter.rs
+++ b/src/getter.rs
@@ -434,6 +434,10 @@ impl Getter for CppCode {
                                 | Cpp::FieldIdentifier
                                 | Cpp::DestructorName
                                 | Cpp::OperatorName
+                                | Cpp::QualifiedIdentifier
+                                | Cpp::QualifiedIdentifier2
+                                | Cpp::QualifiedIdentifier3
+                                | Cpp::QualifiedIdentifier4
                                 | Cpp::TemplateFunction
                                 | Cpp::TemplateMethod => {
                                     let code = &code[first.start_byte()..first.end_byte()];
--- a/src/spaces.rs
+++ b/src/spaces.rs
@@ -378,3 +378,24 @@ impl Callback for Metrics {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::{check_func_space, CppParser};
+
+    #[test]
+    fn c_scope_resolution_operator() {
+        check_func_space::<CppParser, _>(
+            "void Foo::bar(){
+                return;
+            }",
+            "foo.c",
+            |func_space| {
+                insta::assert_json_snapshot!(
+                    func_space.spaces[0].name,
+                    @r###""Foo::bar""###
+                );
+            },
+        );
+    }
+}
--- a/src/tools.rs
+++ b/src/tools.rs
@@ -379,10 +379,10 @@ pub(crate) fn intense_color(stdout: &mut StandardStreamLock, color: Color) -> st
 }
 
 #[cfg(test)]
-pub(crate) fn check_metrics<T: crate::ParserTrait>(
+pub(crate) fn check_func_space<T: crate::ParserTrait, F: Fn(crate::FuncSpace)>(
     source: &str,
     filename: &str,
-    check: fn(crate::CodeMetrics) -> (),
+    check: F,
 ) {
     let path = std::path::PathBuf::from(filename);
     let mut trimmed_bytes = source.trim_end().trim_matches('\n').as_bytes().to_vec();
@@ -390,7 +390,16 @@ pub(crate) fn check_metrics<T: crate::ParserTrait>(
     let parser = T::new(trimmed_bytes, &path, None);
     let func_space = crate::metrics(&parser, &path).unwrap();
 
-    check(func_space.metrics)
+    check(func_space)
+}
+
+#[cfg(test)]
+pub(crate) fn check_metrics<T: crate::ParserTrait>(
+    source: &str,
+    filename: &str,
+    check: fn(crate::CodeMetrics) -> (),
+) {
+    check_func_space::<T, _>(source, filename, |func_space| check(func_space.metrics))
 }
 
 #[cfg(test)]


michaelvanstraten:
@marco-c?

michaelvanstraten:
The test failure is just there being an old snapshot which doesn't capture this problem. I've attached a new snapshot in this comment.
[alphabet.cc.snap.new.txt](https://github.com/mozilla/rust-code-analysis/files/13068319/alphabet.cc.snap.new.txt)

marco-c:
Thanks for the PR! Could you add a test?

michaelvanstraten:
Sorry, could you clarify? 

Should I update the old snapshot and add a unit test, or should I just update the snapshot?

marco-c:
I was asking to add a unit test (in addition to updating the snapshot).
The snapshot tests are nice, but it takes longer to analyze their results compared to a unit test.

michaelvanstraten:
In what file would this go?

marco-c:
Yeah, the function module sounds reasonable. Or really any module for any metric which is affected by the change (like src/metrics/loc.rs?).

michaelvanstraten:
This shouldn't really effect any metric test since the only thing that changes it the `name` in `FuncSpace`.

michaelvanstraten:
Is there a faster way to updated all the snapshot rather than just running
```bash
cargo test --no-fail-fast; cargo insta accept
```
a billion times?

marco-c:
@Luni-4 could you help? I've never had to do it myself.

lu-zero:
I guess you are looking for
```
cargo insta test --accept
```

See the [manual](https://insta.rs/docs/cli/) for the other options :)

michaelvanstraten:
> See the [manual](https://insta.rs/docs/cli/) for the other options :)

Thank you <3

michaelvanstraten:
Okay, now all that's left it to make a PR over at https://github.com/SoftengPoliTo/rca-output.

marco-c:
I moved the https://github.com/SoftengPoliTo/rca-output repo to https://github.com/mozilla/rust-code-analysis-output

michaelvanstraten:
Is the git submodule ref already updated in this repo?

marco-c:
> Is the git submodule ref already updated in this repo?

Not yet, can you do it as part of this PR?

michaelvanstraten:
Okay, I tested this on two Apple Silicon machines and both ran fine. I also tried it on a x64-server and there it complains about two rows in the snapshot being mixed up.

Luni-4:
I guess we should update the `rca-output` submodule in this directory too, so removing this commit [dec5298](https://github.com/mozilla/rust-code-analysis-output/commit/dec52988fb0436b8741b786eee4321838558cde6) and CI should work because it downloads the repository at the right SHA

michaelvanstraten:
I could not reproduce this on my Apple Silicon machines even when following the exact same execution order from the CI workflow. This is something to do with platform specificity @marco-c can you help?

Luni-4:
@michaelvanstraten 

CI uses commit SHA `Submodule path 'tests/repositories/rca-output': checked out '65f11a96bbbe69a6ebaa2a083623cc6bc85a48a9'` while `rca-output` last commit is `dec52988fb0436b8741b786eee4321838558cde6`

michaelvanstraten:
Did Marco rerun the ci?

Luni-4:
@marco-c Can you re-run CI?

marco-c:
Sorry I had missed your message. Rerunning CI wouldn't help, it is not a temporary issue: this PR is updating rca-output to 65f11a96bbbe69a6ebaa2a083623cc6bc85a48a9 and not to dec52988fb0436b8741b786eee4321838558cde6, so even if we rerun the CI, it will still clone 65f11a96bbbe69a6ebaa2a083623cc6bc85a48a9.

marco-c:
So @michaelvanstraten, I think the fix is simply to update the submodule as part of this PR :)

michaelvanstraten:
The submodule now, gets checked out at the right commit and the pipeline still fails. 

I have tested this on three machines by now and on every arm system the tests pass and on x64 they fail.

marco-c:
Maybe there's some approximation somewhere that is different between arm and x86. @Luni-4 any thoughts?

michaelvanstraten:
Okay, I have tested this again on another system which also has an Apple Silicon chip and there the tests also fail.

michaelvanstraten:
To reproduce this, follow these steps:

```bash
# Clone the repository and switch to the specific commit
git clone --quiet https://github.com/michaelvanstraten/rust-code-analysis
cd rust-code-analysis
git -c advice.detachedHead=false checkout e7076d8a90af0bbb48515ed34b23639e99d4f403
git submodule update --init
```

```bash
# Run tests using Rust 1.72
cargo +1.72 test --workspace --verbose --all-features
```
```bash
# Attempt to run tests using Rust 1.71.1 (this fails)
cargo +1.71.1 test --workspace --verbose --all-features
```

Luni-4:
Ouch, this is bad. 

@michaelvanstraten 
Can we update the snapshots and also `insta` to overcome this problem?

michaelvanstraten:
> Can we update the snapshots and also `insta` to overcome this problem?

To what version, to fit the old `rustc` or the up-to-date version?

Luni-4:
> > Can we update the snapshots and also `insta` to overcome this problem?
> 
> To what version, to fit the old `rustc` or the up-to-date version?

We use `rust:buster` as docker image for our CI which implements 1.74 as stable rust version. In my opinion, we should stay with that version.

michaelvanstraten:
Okay, then i will downgrade the snapshots.

Luni-4:
To be clearer, we should stay with Rust stable version and `insta` version that runs on stable. In our case, the most updated `insta` version runst on stable. So we just have to update snapshots using the last stable Rust version and everything should be ok

michaelvanstraten:
Every time I change the rustc version, some line numbers get moved in the snapshot and that trips up insta.
I will try to push update the submodule again for rustc version`1.74.1`, if that doesn't work, ill give up.

Luni-4:
@michaelvanstraten  and @marco-c 

Is the last update worked?

marco-c:
We need to test. @michaelvanstraten can you update the PR to the latest revision of https://github.com/mozilla/rust-code-analysis-output?

michaelvanstraten:
Btw, the new snapshots that work on my machine at home don't work for my work machine (same rust version).

michaelvanstraten:
@marco-c or @Luni-4 can one of you please verify or dismiss that the snapshots are not working for your machines?

Luni-4:
@michaelvanstraten 

I'm going to try locally if I can reproduce the problem. Could you try as last attempt to update `insta`?

Luni-4:
Locally I get the same error that we can see in CI pulling your branch @michaelvanstraten

michaelvanstraten:
@Luni-4, should I update to `1.34.0` the next major version?

Luni-4:
@michaelvanstraten I'm on `Linux`. 

Looking at the differences of my last update, it seems we have a sorting problems on lines https://github.com/mozilla/rust-code-analysis-output/commit/9abe8e86cd97c22bfc4cf7249bac231de1fc6b47

michaelvanstraten:
@Luni-4 I mentioned that in the PR to merge changes, the position changes in the list should not be considered a failure.

michaelvanstraten:
Or should we sort lines?

Luni-4:
`Ploc` lines are saved in a `FxHashet`, so the serialization happens in an arbitrary order. Until now we haven't changed the snapshots, so the problem did not happen. 
`insta` does not sort out things unless it is specified. So we need to add 

```rust
// Sort hashmaps to avoid having different orders in snapshots
sort_maps => true,
```
 down this [line](https://github.com/mozilla/rust-code-analysis/blob/master/tests/common/mod.rs#L60).

Could you update your PR @michaelvanstraten and see if it works?

Luni-4:
Another solution could be to compare `json` snapshots, as we already do with unit tests, not to worry about arbitrary serialization. In this way we do not need to sort out things. We need to change [this](https://github.com/mozilla/rust-code-analysis/blob/master/tests/common/mod.rs#L58-L64) though

michaelvanstraten:
I would go with the easy solution of sorting the lists, but then we would have to update the submodule again!

michaelvanstraten:
Are you sure the `sort_maps` will sort the `HashSet`?

Luni-4:
Let's try to refresh the `snapshots` with the ones just committed. If that does not work, let's add the `sort_maps` change

michaelvanstraten:
Before I update the snapshots again, could you check that with `sort_maps` the lines are not sorted?

michaelvanstraten:
I mean the new snapshots would just revert to, [dec52988](https://github.com/mozilla/rust-code-analysis-output/commit/dec52988fb0436b8741b786eee4321838558cde6).

Luni-4:
> I mean, the new snapshots would just revert to, [dec52988](https://github.com/mozilla/rust-code-analysis-output/commit/dec52988fb0436b8741b786eee4321838558cde6).

This is the last commit https://github.com/mozilla/rust-code-analysis-output/commit/6fc922450a5d8c8339eb8449fd4110eb10042d76

Luni-4:
@michaelvanstraten 

I've finally found the problem: we did not serialize into `json` files our data, so we have discrepancies among data we wanted to show and the real data returned by the `FuncSpace` struct. If you apply this [change](https://github.com/mozilla/rust-code-analysis/pull/1080/commits/a510f868d1ec9aef66a7ea66f7d4cd25f1a15829) to your code, I can produce new snapshots and we are ready to go

michaelvanstraten:
> You can remove the last commit and then update to the last `rca-output` commit and it should work ( 🤞 )

What I was trying to say in the comment about updating the snapshots is that the update would just revert the `rust-code-analysis-output` to [dec52988](https://github.com/mozilla/rust-code-analysis-output/commit/dec52988fb0436b8741b786eee4321838558cde6). A new commit would be added, yes, but the diff between the new commit and the one mentioned would be zero.

michaelvanstraten:
> I can produce new snapshots and we are ready to go

Maybe while you're at it, squash the ones that failed?

michaelvanstraten:
> This is the last commit [mozilla/rust-code-analysis-output@6fc9224](https://github.com/mozilla/rust-code-analysis-output/commit/6fc922450a5d8c8339eb8449fd4110eb10042d76)

Sorry, for the confusion, I didn't realize that you already updated the snapshots.

I tried it with the serialization logic and now the only problem left is that the absolute [`file paths`](https://github.com/mozilla/rust-code-analysis-output/blob/ec86960bdcb912bd627320d012cd97414619b518/snapshots/pdf.js/src/pdf.js.snap#L6) are serialized.

michaelvanstraten:
Snapshot [`fix`](https://github.com/mozilla/rust-code-analysis/pull/1081) would be ready.

Luni-4:
@michaelvanstraten 

Thanks for your work!

In my opinion we can:

1. Update submodules in this PR
    1. Run `git submodule update --remote tests/repositories/rca-output`
    2. Commit `tests` directory changes
    3. Push
 2. Merge this PR
 
 Then we can work on the other PR to fix json problems. What do you think? @marco-c any observation about this plan?

michaelvanstraten:
> Then we can work on #1081 to fix json problems.

My main problem with this that we now incorperate a temporary fix that indroduces a lot of new data into the `rca-output` repo which then is useless then we apple the fix.

Luni-4:
This is the last commit of `rca-output` https://github.com/mozilla/rust-code-analysis-output/commit/f8b7e612ad61c14699c8b3a2fc9e62f56f6d0f2a

So we do not add anything different with this PR, just a simple difference with the `json` files

michaelvanstraten:
@Luni-4 can you update the snapshots with this:

```rust
let value = serde_json::ser::to_string_pretty(&FuncSpace {
        name: None, // !
        ..get_function_spaces(&language, source, &path, None).unwrap()
    })
    .unwrap();
```

They currently still include the absolute path.

Luni-4:
@michaelvanstraten 

Sure, thank you for spotting that! If you add to this PR, I'm going to create new snapshots on `rca-output` starting from that

michaelvanstraten:
@Luni-4, should be added.

Luni-4:
Done https://github.com/mozilla/rust-code-analysis-output/commit/1da112c4d2487e8e0f53fa7781006f13ac6ff0ea

You can update `rca-output` submodule now

michaelvanstraten:
The float values are not rounded, that's exactly why I first wanted to merge #1081.

Luni-4:
Yes, right, but let's merge this one first and then we fix the rounded problem in that PR. Right now the CI is broken, so still leaving that broken is not a problem

michaelvanstraten:
> Right now the CI is broken, so still leaving it broken is not a problem

Okay, can I then drop the commits addressing the CI issue?

Luni-4:
Yes, we can move every CI commit in the other PR

michaelvanstraten:
> Yes, we can move every CI commit in the other PR

Sorry for the long wait, it was christmas. The CI commits have been removed.

Luni-4 (approved):
Do not worry!  Thanks a lot! 

@marco-c  Can we merge this PR?

PR #1080, opened 2023-12-19
Replace `fxhash` with `std::collections::HashMap`

Luni-4:
This PR replaces `fxhash` dependency, which is now deprecated since its last release is more than 5 years ago, with `std::collections::HashMap`

### DIFF
--- a/src/metrics/cognitive.rs
+++ b/src/metrics/cognitive.rs
@@ -1,4 +1,5 @@
-use fxhash::FxHashMap;
+use std::collections::HashMap;
+
 use serde::ser::{SerializeStruct, Serializer};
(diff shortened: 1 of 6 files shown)


src/metrics/cognitive.rs
```diff
@@ -1755,7 +1756,7 @@ mod tests {
     fn java_single_branch_function() {
         check_metrics::<JavaParser>(
             "class X {
-                public static void print(boolean a){  
+                public static void print(boolean a){
```
marco-c, line 1759: Could you revert these unrelated whitespace changes?

marco-c (approved):
Thanks!

PR #1081, opened 2023-12-20
Modify tests for more deterministic behavior

michaelvanstraten:


Luni-4:
@michaelvanstraten 

Now that #1078 has been merged, we can fix CI using this PR. Let's use `yaml` comparisons and not `json` because of the reccomandation written [here](https://docs.rs/insta/1.34.0/insta/macro.assert_json_snapshot.html).

We need to update the snapshots accordingly.

michaelvanstraten:
> Let's use `yaml` comparisons

Already implemented, for the same reason.

> We can also integrate the missing features present in #1080

Do you mean the switch to `std::collections::Hashmap`, the sorting of maps or both?

Luni-4:
We can just sort out the maps in this PR, thanks for the catch!

michaelvanstraten:
@Luni-4, tests pass except for the three sitter checks, can you check it out?

Luni-4 (changes requested):
Thanks a lot! For what concerns `tree-sitter` checks, it seems it's running them because of some changed values. We can leave them as-is for now

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "tests/repositories/rca-output"]
 	path = tests/repositories/rca-output
-	url = https://github.com/mozilla/rust-code-analysis-output
+	url = https://github.com/michaelvanstraten/rust-code-analysis-output
```
Luni-4, line 3: We have to revert this change

Luni-4 (changes requested):

.gitmodules
michaelvanstraten: https://github.com/mozilla/rust-code-analysis-output/pull/4

Luni-4: Merged, thanks! Let's update this PR

Luni-4 (approved):
Thank you @michaelvanstraten! 

@marco-c In my opinion we can merge this PR and open an issue to find the possible issues with grammars

Issue #1091, opened 2024-08-06
Fix test failures introduced by #1086


marco-c:

marco-c:
@alexle0nte are you going to work on this?

alexle0nte:
Yes, I'll create a PR to fix all the tests

marco-c:
@alexle0nte are you making progress on this?

alexle0nte:
@marco-c Yes, #1125 fixes all the metrics tests. 
However, there is a problem with repository tests, whose snapshots are contained in `rca-output`.

Version `0.21.2` of `tree-sitter-rust` has a [bug](https://github.com/tree-sitter/tree-sitter-rust/issues/236) regarding parsing doc comments, that causes many `serde` tests to fail. 
This bug has been fixed in version `0.23.0`, however, upgrading `tree-sitter-rust` to this version would also require upgrading all other grammars to `0.23.0`, but `tree-sitter-kotlin` currently [doesn't yet support `tree-sitter 0.23.0`](https://github.com/fwcd/tree-sitter-kotlin/issues/148).

marco-c:
@alexle0nte do you think we could submit a PR to tree-sitter-kotlin to support 0.23.0 or is it hard?

Your plan seems fine otherwise.

alexle0nte:
@marco-c I've just opened a PR in `tree-sitter-kotlin` to bump the `tree-sitter` dependency to `23.0`. Once it's merged and the new release is out, we can upgrade all the grammars to `23.0`.

marco-c:
Thanks! If it still open in a couple of week, we could create a temporary fork to unblock the updates.

Luni-4:
@marco-c and @alexle0nte 

We can also ask `tree-sitter` developers if they are willing to adopt the `Kotlin` grammar among their organization's parsers

marco-c:
@Luni-4 @alexle0nte it looks like the PR is still not merged, wdyt we should do?

alexle0nte:
> @Luni-4 @alexle0nte it looks like the PR is still not merged, wdyt we should do?

I think we can create a temporary fork as you suggested and use that until it gets merged

marco-c:
We can close this now, right @alexle0nte ?

alexle0nte:
> We can close this now, right [@alexle0nte](https://github.com/alexle0nte) ?

Yes

PR #1113, opened 2024-09-08
rust: a question mark operator is a conditional exit

samueltardieu:
A question mark operator `?` may exit the current function the same way a `return`
will. It must be counted in the "nexits" statistics.

### DIFF
--- a/src/metrics/exit.rs
+++ b/src/metrics/exit.rs
@@ -156,7 +156,7 @@ impl Exit for TsxCode {
 
 impl Exit for RustCode {
     fn compute(node: &Node, stats: &mut Stats) {
-        if matches!(node.kind_id().into(), Rust::ReturnExpression)
+        if matches!(node.kind_id().into(), Rust::ReturnExpression | Rust::QMARK)
             || Self::is_func(node) && node.child_by_field_name("return_type").is_some()
         {
             stats.exit += 1;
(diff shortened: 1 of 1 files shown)


samueltardieu:
Is there something missing for this PR to be reviewed and merged?

calixteman (approved):
LGTM. Thank you.

alexle0nte:
While working on #1091, I noticed that some tests are failing because all nodes corresponding to a `Rust::QMARK` are mistakenly counted as exit points.

For example, in this code:
```rust
pub fn assert_ser_tokens<T: ?Sized>(value: &T, tokens: &[Token])
where
    T: Serialize,
{
    let mut ser = Serializer::new(tokens);
    match value.serialize(&mut ser) {
        Ok(_) => {}
        Err(err) => panic!("value failed to serialize: {}", err),
    }

    if ser.remaining() > 0 {
        panic!("{} remaining tokens", ser.remaining());
    }
}
```
the `?` on the first line, inside `<T: ?Sized>`, should not be counted as an exit point. 
Therefore, adding `Rust::QMARK` in an OR condition with `Rust::ReturnExpression` doesn't fully address the issue.

calixteman:
@alexle0nte could you file a bug please ?
@samueltardieu would you mind to have a look on it ?

Thank you.

PR #1124, opened 2024-09-24
rust: distinguish a question mark operator from a trait marker

### LINKED ISSUE #1123: Fix handling of `?` operator in Rust `nexits` metric
As pointed out by #1113, the `?` operator can also represent an exit point within a Rust function, and therefore, it must be counted in the `nexits` metric.

However, not all `Rust::QMARK` nodes should be counted as exit points.
For example, in this code:
```rust
pub fn assert_ser_tokens<T: ?Sized>(value: &T, tokens: &[Token])
where
    T: Serialize,
{
    let mut ser = Serializer::new(tokens);
    match value.serialize(&mut ser) {
        Ok(_) => {}
        Err(err) => panic!("value failed to serialize: {}", err),
    }

    if ser.remaining() > 0 {
        panic!("{} remaining tokens", ser.remaining());
    }
}
```
the `?` on the first line, inside `<T: ?Sized>`, doesn't represent an exit point.

samueltardieu:
Fix #1123

### DIFF
--- a/src/metrics/exit.rs
+++ b/src/metrics/exit.rs
@@ -156,8 +156,10 @@ impl Exit for TsxCode {
 
 impl Exit for RustCode {
     fn compute(node: &Node, stats: &mut Stats) {
-        if matches!(node.kind_id().into(), Rust::ReturnExpression | Rust::QMARK)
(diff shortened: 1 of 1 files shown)


marco-c (approved):
Thanks!

PR #1141, opened 2025-03-28
Update grammars

alexle0nte:
This PR updates `tree-sitter` and grammars to the latest versions.


alexle0nte:
@marco-c @Luni-4 before merging this, I think we have to first merge mozilla/rust-code-analysis-output#5

marco-c:
@alexle0nte merged it!

alexle0nte:
> @alexle0nte merged it!

Thanks! I synced the submodule. Now it's ready for review

marco-c:
There are a few test failures, for example https://community-tc.services.mozilla.com/tasks/Dk7Mn21LRpWpwePp_KHCew/runs/0/logs/public/logs/live.log.

alexle0nte:
> There are a few test failures, for example https://community-tc.services.mozilla.com/tasks/Dk7Mn21LRpWpwePp_KHCew/runs/0/logs/public/logs/live.log.

Yes, this test fails due to a parsing error in tree-sitter-cpp. I have opened an issue to report it: tree-sitter/tree-sitter-cpp#307

marco-c:
Same for https://community-tc.services.mozilla.com/tasks/ViJGwdjsR-2f322NYxBvSQ/runs/0/logs/public/logs/live.log?

alexle0nte:
Yes, all these pdf.js tests fail due to the same parsing error in tree-sitter-javascript. As I was suggesting the other day, after merging this PR, I can open issues to report the failing tests and their causes, linking the corresponding grammar parsing error issue for each of them

marco-c:
OK, thanks. Just wanted to confirm these were all known!

marco-c (approved):

marco-c:
Looks good to me overall. @Luni-4 do you want to take a look too?

Luni-4 (approved):
Thanks a lot for your hard work @alexle0nte! This is fine for me too! 

Just a question which can treat in another PR perhaps: How can we make the CI green? Since tests do not pass, now it is impossible to publish a new version of rca for example.

alexle0nte:
> Thanks a lot for your hard work @alexle0nte! This is fine for me too! 
> 
> 
> 
> @alexle0nte and @marco-c 
> 
> 
> 
> Just a question we can treat in another PR perhaps: How can we make the CI green? Since tests do not pass, now it is impossible to publish a new version of rca for example.

I think we can temporarily ignore the failing tests and add a comment explaining the reason, as well as a link to the issue we will open.

Luni-4:
I agree with this plan

marco-c:
It would be ideal to mark the failing tests as failing, so the CI is green and we notice if there is a new failure in addition to the known ones (otherwise the new failures will be lost in the noise).
We can do this in a follow-up.

marco-c:
@alexle0nte could you file issues for the failing tests and link to the upstream grammar issues?

Issue #1143, opened 2025-03-31
Fix `tree-sitter-mozjs` test failures introduced by #1141


alexle0nte:
Following the grammar updates made by #1141, some tests are failing due to a parsing error in `tree-sitter-javascript`: tree-sitter/tree-sitter-javascript#351. 
Below is a list of the failing tests, intended to track and fix them as the issue is resolved:
- `pdf.js/extensions/firefox/tools/l10n.js`
- `pdf.js/src/display/optional_content_config.js`
- `pdf.js/docs/contents/js/bootstrap.min.js`
- `pdf.js/src/display/font_loader.js`
- `pdf.js/src/display/content_disposition.js`
- `pdf.js/src/display/pattern_helper.js`
- `pdf.js/src/display/canvas.js`
- `pdf.js/src/display/annotation_layer.js`
- `pdf.js/src/display/svg.js`
- `pdf.js/docs/contents/js/jquery-2.1.0.min.js`
- `pdf.js/src/display/display_utils.js`
- `pdf.js/src/display/editor/ink.js`
- `pdf.js/src/display/text_layer.j`s
- `pdf.js/src/display/api.js`
- `pdf.js/src/core/fonts_utils.js`
- `pdf.js/src/core/primitives.js`
- `pdf.js/src/core/jpg.js`
- `pdf.js/src/core/jbig2_stream.js`
- `pdf.js/src/core/colorspace.js`
- `pdf.js/src/core/worker.js`
- `pdf.js/src/core/parser.js`
- `pdf.js/src/core/annotation.js`
- `pdf.js/src/core/evaluator.js`
- `pdf.js/src/core/function.js`
- `pdf.js/src/core/jpx_stream.js`
- `pdf.js/src/core/predictor_stream.js`
- `pdf.js/src/core/run_length_stream.js`
- `pdf.js/src/core/image.js`
- `pdf.js/src/core/writer.js`
- `pdf.js/src/core/operator_list.js`
- `pdf.js/src/core/jbig2.js`
- `pdf.js/src/core/jpx.js`
- `pdf.js/src/core/document.js`
- `pdf.js/src/core/name_number_tree.js`
- `pdf.js/src/core/cff_parser.js`
- `pdf.js/src/core/core_utils.js`
- `pdf.js/src/core/metadata_parser.js`
- `pdf.js/src/core/pattern.js`
- `pdf.js/src/core/fonts.js`
- `pdf.js/src/core/ccitt.js`
[...]

Issue #1142, opened 2025-03-31
Fix `tree-sitter-cpp` test failures introduced by #1141


alexle0nte:
Following the grammar updates made by #1141, some tests are failing due to parsing errors in `tree-sitter-cpp`.
Below is a list of the failing tests and their corresponding upstream issues, which track the errors so that the tests can be fixed as the issues are resolved:
- `test_fn_id_strings`: [#307](https://github.com/tree-sitter/tree-sitter-cpp/issues/307)
- `DeepSpeech/native_client/deepspeech.cc`: [#308](https://github.com/tree-sitter/tree-sitter-cpp/issues/308)
- `DeepSpeech/native_client/getopt_win.h`: [#308](https://github.com/tree-sitter/tree-sitter-cpp/issues/308)
- `DeepSpeech/native_client/kenlm/util/mmap.cc`: [#308](https://github.com/tree-sitter/tree-sitter-cpp/issues/308)
- `DeepSpeech/native_client/deepspeech.h`: [#309](https://github.com/tree-sitter/tree-sitter-cpp/issues/309) and [#310](https://github.com/tree-sitter/tree-sitter-cpp/issues/310)
- `DeepSpeech/native_client/kenlm/util/double-conversion/fast-dtoa.cc`: [#311](https://github.com/tree-sitter/tree-sitter-cpp/issues/311)
- `DeepSpeech/native_client/kenlm/lm/left_test.cc`: [#312](https://github.com/tree-sitter/tree-sitter-cpp/issues/312)
- `DeepSpeech/native_client/ctcdecode/third_party/openfst-1.6.7/src/test/fst_test.h`: [#252](https://github.com/tree-sitter/tree-sitter-cpp/issues/252)
- `DeepSpeech/native_client/ctcdecode/third_party/openfst-1.6.9-win/src/include/fst/test/fst_test.h`: [#252](https://github.com/tree-sitter/tree-sitter-cpp/issues/252)

PR #1144, opened 2025-03-31
Mark and ignore failing tests to pass CI

alexle0nte:
This PR marks and temporarily ignores some failing tests due to parsing errors in the upstream grammars (see issues #1142 and #1143), allowing the CI to pass.


marco-c (approved):
Thanks!

PR #1155, opened 2025-04-24
Update book documentation and add deploy on push to master

alexle0nte:


Luni-4 (approved):
This is fine for me! Thanks a lot!

In this way, the documentation is always updated at each push on `master`, without waiting for to schedule a new release. Let's try out if it works!

PR #1158, opened 2025-05-08
Exclude internal grammars from language stats

alexle0nte:


Luni-4 (approved):
This is fine, thanks a lot! Now `rca` becomes a real Rust crate :smile:

PR #1165, opened 2025-05-16
Bump `tree-sitter` and `tree-sitter-rust`

alexle0nte:


marco-c:
It looks like `check tree-sitter-mozcpp` took too long. We might need to increase its maximum runtime

marco-c:
Even with a maximum runtime of 3 hours, it's still running out of time. Have you tried running it locally? I wonder if there's a deadlock somewhere, or why it got so much slower than it was before.

marco-c:
This was done in #1207.

PR #1213, opened 2026-01-20
feat(node): Add utility methods for AST traversal

marlon-costa-dc:
## Summary

This PR adds utility methods to the `Node` struct and `Search` trait for improved AST traversal capabilities:

- **`has_ancestor()`** - Check if any ancestor node matches a predicate
- **`all_occurrences()`** - Find all descendant nodes matching a predicate (extends `Search` trait)

## Origin

These methods were developed to support more complex AST analysis patterns, particularly for:
- Checking node context by examining ancestors (e.g., "is this node inside a class?")
- Collecting all nodes of a certain type for batch processing

### `has_ancestor<F: Fn(&Node) -> bool>(&self, pred: F) -> bool`

Traverses up the tree from the current node's parent to the root, returning `true` if any ancestor satisfies the given predicate.

```rust
// Example: Check if node is inside a function
let in_function = node.has_ancestor(|n| n.kind_id() == FunctionDeclaration);
```

### `all_occurrences(&self, pred: fn(u16) -> bool) -> Vec<Node<'a>>`

Starting from the current node, finds all descendant nodes (including self) that match the predicate. Uses depth-first traversal.

```rust
// Example: Find all function calls in a subtree
let calls = node.all_occurrences(|id| id == CallExpression);
```

## Implementation decisions

1. **`has_ancestor` uses a generic closure** (`F: Fn(&Node) -> bool`) instead of `fn(u16) -> bool` to allow more flexible predicates that can inspect the full node, not just the kind_id.

2. **`all_occurrences` uses `fn(u16) -> bool`** to match the existing `first_occurrence` signature in the `Search` trait, maintaining API consistency.

3. **Both methods are non-recursive** using explicit stacks to avoid stack overflow on deeply nested ASTs.

## Test plan

- [x] Code compiles without errors
- [x] Existing tests pass
- [ ] New methods are ready for use by downstream implementations

---

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/node.rs
+++ b/src/node.rs
@@ -183,6 +183,21 @@ impl<'a> Node<'a> {
         }
         res
     }
+
+    /// Checks if this node has any ancestor that meets the given predicate.
+    ///
+    /// Traverses up the tree from this node's parent to the root,
+    /// returning true if any ancestor satisfies the predicate.
+    pub fn has_ancestor<F: Fn(&Node) -> bool>(&self, pred: F) -> bool {
+        let mut node = *self;
+        while let Some(parent) = node.parent() {
+            if pred(&parent) {
+                return true;
+            }
+            node = parent;
+        }
+        false
+    }
 }
 
 /// An `AST` cursor.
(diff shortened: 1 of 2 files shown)


marlon-costa-dc:
Replaced by #1224 (clean single-commit branch).

PR #1224, opened 2026-02-25
feat(node): Add AST traversal utilities and expose Node API

marlon-costa-dc:
## Summary

Adds utility methods for AST traversal to the `Node` struct and `Search` trait:

- **`has_ancestor(pred)`** — Check if any ancestor node matches a predicate (non-recursive, explicit stack)
- **`traverse_children(token_list)`** — Navigate multi-level child paths for finding deeply nested nodes
- **`all_occurrences(pred)`** — Find all descendant nodes matching a predicate (depth-first, extends `Search` trait)
- **`asttools` module** — Reusable AST traversal utilities (exported as `pub mod`)
- **`pub` visibility** — Expose `Node` methods (`kind_id`, `parent`, `start_row`, etc.) for downstream crate usage
- **CppCode getter fix** — Fix let-chain pattern to use proper `if let` with `&&`

## Motivation

These methods support more complex AST analysis patterns needed by downstream crates that use `rust-code-analysis` as a dependency:
- Checking node context by examining ancestors (e.g., "is this node inside a class?")
- Collecting all nodes of a certain type for batch processing
- Navigating deeply nested AST structures (e.g., finding visibility modifiers through modifier lists)

## Implementation Notes

- `has_ancestor` uses a generic closure (`Fn(&Node) -> bool`) for flexible predicates
- `all_occurrences` uses `fn(u16) -> bool` to match existing `first_occurrence` signature
- All traversals are non-recursive (explicit stacks) to avoid stack overflow on deep ASTs
- Single clean commit rebased on upstream/master

## Replaces

Replaces #1213 (closed due to messy branch history). Clean single-commit version.

## Test Plan

- [x] `cargo check` passes
- [x] All 252 existing tests pass (`cargo test --workspace`)

### DIFF
--- a/src/getter.rs
+++ b/src/getter.rs
@@ -438,31 +438,30 @@ impl Getter for CppCode {
                     return std::str::from_utf8(code).ok();
                 }
                 // we're in a function_definition so need to get the declarator
-                if let Some(declarator) = node.child_by_field_name("declarator") {
-                    let declarator_node = declarator;
-                    if let Some(fd) = declarator_node.first_occurrence(|id| {
+                if let Some(declarator) = node.child_by_field_name("declarator")
+                    && let Some(fd) = declarator.first_occurrence(|id| {
                         Cpp::FunctionDeclarator == id
                             || Cpp::FunctionDeclarator2 == id
                             || Cpp::FunctionDeclarator3 == id
-                    }) && let Some(first) = fd.child(0)
-                    {
-                        match first.kind_id().into() {
-                            Cpp::TypeIdentifier
-                            | Cpp::Identifier
-                            | Cpp::FieldIdentifier
-                            | Cpp::DestructorName
-                            | Cpp::OperatorName
-                            | Cpp::QualifiedIdentifier
-                            | Cpp::QualifiedIdentifier2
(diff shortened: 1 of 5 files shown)


Luni-4:
@marlon-costa-dc 

This [CI error](https://community-tc.services.mozilla.com/tasks/dieYo3wESXSazmWAK-__oA/runs/0/logs/public/logs/live.log#L1181) should be fixed in another PR, so not blocking for these contributions

src/getter.rs
```diff
@@ -438,31 +438,30 @@ impl Getter for CppCode {
...
+                        | Cpp::QualifiedIdentifier
+                        | Cpp::QualifiedIdentifier2
+                        | Cpp::QualifiedIdentifier3
+                        | Cpp::QualifiedIdentifier4
+                        | Cpp::TemplateFunction
+                        | Cpp::TemplateMethod => {
+                            let code = &code[first.start_byte()..first.end_byte()];
+                            return std::str::from_utf8(code).ok();
                         }
+                        _ => {}
```
Luni-4, line 464: Can these changes be moved in another PR?

Luni-4 (changes requested):

